PR #59, opened 2014-05-07
Resender

RPeschke:
Resender:

This class acts like a normal Producer but as input it takes RAW files. 

Since it has no control over runcontrol the Configuration file will not be resent. From runcontrol side you have to configure before you can run. This will be used as configuration for the new run. For now I don’t see how one could fix this without making the Runcontrol part of the Resender.


simonspa:
Closing this pull request as it is superseded by #102

PR #93, opened 2014-09-29
Move plugins into their producer's source directories

simonspa:
This PR tries to clean up eudaq's source structure a bit and simplify compilation:

* Move Data converter plugins into the producer directory they belong to
* Collect the source files from all /built/ producers and compile them into the main library
* Current behaviour is preserved: all converter plugins in `main/lib/plugins` are included for compilation.

Benefits:
* It is possible to have converters with producer-specific includes (external decoding libraries etc) without affecting eudaq
* at compile time only the converters necessary are built.

Drawbacks:
* when changing the build configuration, also the main library has to be rebuilt because another converter is added/removed.
* this setup does not work if several producer use the same converter plugin. But since the old behaviour is still present and working, placing those converters in `main/lib/plugins` works just fine.

Todo:
* there are still some converters left I couldn't figure out where they belong to.


simonspa:
I'm not sure it is a good idea to follow this approach since it adds quite an ugly CMake overhead. While I still see the need for getting the converters out the main directory (also due to possible external dependencies of the producers) I don't think this is the right approach.

Closing this PR for now.

PR #111, opened 2014-11-29
merging MultipleDataCollectors options into v1.4-dev

rubinsky:
extending "native2" data type to match multiple data collectors DAQ scheme.
- additing ExampleTLUProducer (sends TLUEvents with system clocks for timestamps)
- changes to ExampleProducers - only ZeroSuppressed data sending (RAW commented out till next revision)
- factorisation of STARTRUN scripts
- bug fix in terminating GUI LogCollector

### DIFF
--- a/STARTRUN
+++ b/STARTRUN
@@ -1,1 +1,1 @@
-etc/scripts/STARTRUN.local
+etc/scripts/STARTRUN
--- a/conf/ExampleConfig.conf
+++ b/conf/ExampleConfig.conf
@@ -5,15 +5,11 @@
 RunSizeLimit = 1000000000
 
 [DataCollector]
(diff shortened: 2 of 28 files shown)


simonspa:
Browsing through this PR it looks like this is actually more about reworking STARTRUN scripts and adding a new producer called `ExampleTLUProducer`. I don;t see the necessary changes to `eudaq:RunControl` to actually assign the producers to their respective data collector.

Also I guess it would be necessary to iterate once more (offline) through the possible pitfalls of not doing online event building before adding this functionality to eudaq.

simonspa:
Closing this PR. The functionality claimed in the title has been implemented afaik, other changes are not really motivated here. If this code should go in, a new PR that actually merges fine is needed.

PR #112, opened 2014-12-11
miniTLU aidaPacket threaded

fcrescio:
miniTLU producer is now able to deal with single word 64-bit packets and it handles correctly the case when FIFO reads splits a 2-word packet.
Tested with a miniTLU without errors.


simonspa:
This looks fine to me, I will merge the code - even though I'm not entirely sure about the future of the `v2.0-dev` branch. Are these changes protable to the `v1.4-dev` branch? That one is the one which will be running in the test beams this year.

PR #116, opened 2015-02-17
CMSPixel Producer

simonspa:
This pull request adds a new producer for CMS Pixel Detector modules. This includes the producer itself, the converter plugin for eudaq's main library as well as integration into the native reader library for conversion to EUTelescope's LCIO.

Since the converter plugin currently depends on an external library the presence of this library is checked and the plugin is only built when the requirement is satisfied. Otherwise libEUDAQ is built without support for "cmspixel" event type conversion. This should not pose a problem since all users requiring this conversion need the external library anyway, all others don't care.

If however the producer build is requested (via the normal `cmake -DBUILD_cmspixel ..`) and the external library is not present, a fatal error will be thrown.

There is one speciality about this producer: we are running two detectors with exactly the same but independent DAQ. This means we have two producers putting out a data stream with the same event type. This (currently) makes it impossible to distinguish between them when decoding the native raw data - since subevents are only matched agains their event type but no producer ID or similar. This requires some workarounds: we introduced *two* new event types (written by the same producer though) which allow later on when decoding to distinguish between the two detectors. Since the data content is the same, `CMSPixelHelpers.hh` does the actual decoding and `CMSPixel*DecoderPlugin.cc` are only wrappers for this class which are setting the plane ID according to the event type. Not a nice hack, but currently the only way without major restructuring of eudaq's core.

This DAQ ran successfully in two testbeams at DESY, very stably and with online monitoring, correlation plots of the DUT planes and all the bells and whistles.

This pull request also contains a few minor changes to eudaq core components:
 * FindEUTelescope.cmake: just e a bit more verbose, this helps in debugging user compilation problems.
 * STARTRUN.local: check the LD_LIBRARY_PATH before appending path.
 * ATLASFEI4Interpreter: this is a compiler fix by @tbisanz merged into this branch
 * main/lib/CMakeLists.txt: this CMake snippet removes the cmspixel converter in case dependencies are not satisfied
 * main/lib/src/DataCollector.cc: these changes decrease the printouts (only every 1000 events instead of 100 events, should be enough...) and adds an additional printout for every EORE event received from any producer. This is really helpful for debugging problematic setups with triggering problems and does not increase printout much.
 * monitors/onlinemon/include/SimpleStandardEvent.hh: this is a compilation fix for Mac OS X
 *  monitors/onlinemon/src/OnlineMon.cxx: this removes a (in my eyes) useless printout for every event processed.


simonspa:
Anyone has any comments or objections? Otherwise I would like to merge this rather sooner than later...

PR #117, opened 2015-02-25
Swapped SetPMTVcntlMod and SetPMTVcntl

pheloniusfriar:
It's always a good idea to set a required flag before invoking a function that needs it... In this case, the flag indicating whether or not a TLU that has been modified to generate 0->2V Vcntl signals to the PMTs was being set after the Vcntl voltages, so it would fail if using a so-modified TLU. This bug had no impact unless the modified TLU feature was being used.


simonspa:
Hi @pheloniusfriar, the patch looks good and useful. I will merge this next week instead of now, just to remind myself of cherry-picking the commit to the stable maintenance branch v1.4-dev.

PR #127, opened 2015-04-17
Fixes to OnlineMonitor

simonspa:
* Fix crash if nPlanes change between two runs (e.g. added new producer and reconfigured)
  * Fix crash when "AutoReset" was activated before first run was started
  * Add new projected hitmap histograms


simonspa:
One more additional change:

The cmake default build type is now set to "RelWithDebInfo" to have debug information available by default. Before the build type was empty by default, resulting in binaries with no debug symbols.

PR #129, opened 2015-04-22
Root producer

RPeschke:
bugfixes:

[CORE]
allow multiple producer from one program

[ROOTProducer]
removed race condition between set status function and send event function by adding a delay.

### DIFF
--- a/main/include/eudaq/LogSender.hh
+++ b/main/include/eudaq/LogSender.hh
@@ -18,7 +18,7 @@ namespace eudaq {
       LogSender();
       ~LogSender();
       void Connect(const std::string & type, const std::string & name, const std::string & server);
-      void disconnect();
+      void Disconnect();
       void SendLogMessage(const LogMessage &);
       void SetLevel(int level) { m_level = level; }
       void SetLevel(const std::string & level) { SetLevel(Status::String2Level(level)); }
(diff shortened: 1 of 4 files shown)


main/include/eudaq/LogSender.hh
```diff
@@ -17,6 +18,7 @@ namespace eudaq {
       LogSender();
       ~LogSender();
       void Connect(const std::string & type, const std::string & name, const std::string & server);
+      void disconnect();
```
simonspa: Minor style question: should we capitalize this function name as all others are? `Disconnect()`?

main/lib/src/LogSender.cc
```diff
@@ -10,6 +10,12 @@ namespace eudaq {
     m_logclient(0), m_errlevel(Status::LVL_DEBUG), m_shownotconnected(false) {}
 
   void LogSender::Connect(const std::string & type, const std::string & name, const std::string & server) {
+    MutexLock m(m_mutex);
+    if (isConnected)
+    {
```
simonspa: Minor style issue:
in other code around we don't have dangling parentheses, maybe also write
```
if() {
```
instead of
```
if()
{
```

main/lib/src/LogSender.cc
```diff
@@ -43,7 +49,16 @@ namespace eudaq {
     if (std::string(packet, 0, i1) != "OK") EUDAQ_THROW("Connection refused by LogCollector server: " + packet);
   }
 
+
+  void LogSender::disconnect()
+  {
```
simonspa: see above

simonspa:
Two style issues, otherwise fine to be merged from my side.

PR #133, opened 2015-05-15
Add CLICpix converter to support CLICpix in Online Monitor.

adrianf0:


simonspa:
Thanks for contributing back, merged your converter plugin.

PR #135, opened 2015-05-21
Integration of code developed for the ALICE ITS Upgrade

freidt:
Hello,

on behalf of the ALICE ITS Upgrade team I would like to ask for integration of our code into the official repository.

Some of them are very local (producers, converters) but some of the touch the main code.

The following changes to the main code have been implemented:
- Added additional event flags
- Added the possibility to run up to a certain event count instead of only a file size limit
- Using flags to reject events in native reader
- Added possibility to change the configuration file after the first run with this file is done //EDIT

Regards,
Felix
on behalf of the ALICE ITS Upgrade Characterisation Team


simonspa:
Hi @freidt 

thanks for your contribution. I haven't closely looked at the EUDAQ library changes (I will do so) but have already one first comment:

Could you please remove all sort of configuration file and data files from your pull request? I see lots of new files ending with `.txt` or `.conf` that should not go into this repository. Please remove all these files, this helps keeping the code base cleaner.

If you need to keep track of settings and configurations and share them with collaborators I recommend setting up a private dedicated git repository inside the ~./conf/` directory and committing your configurations there.

Cheers,
Simon

freidt:
Hi Simon,

thanks for your very quick reply. Am I understanding you correctly, that you don't want to have those files at all in the repository? Or just not outside of our producer folders? Would it be fine, if we were installing them using CMake during the compilation of our producers?

Cheers,
Felix

simonspa:
Dear @freidt 

I would vote for not installing them at all - because this is data that will be delivered to every single EUDAQ user. CMake would be a nice way to get this sort of dependencies - but please keep it in the CMake file inside your producer's code directories.

Would that be okay with you?

/Simon

freidt:
Hi Simon,

everything is remove from logs/ and conf/
I'll discuss with my colleagues which files we could additionally remove.

Cheers,
Felix

gui/include/euRun.hh
```diff
@@ -31,126 +31,142 @@ inline std::string to_bytes(const std::string & val) {
...
+  typedef std::map<std::string, QLabel *> status_t;
+  status_t m_status;
+  int m_prevtrigs;
+  double m_prevtime, m_runstarttime;
+  long long m_filebytes;
+  long long m_events;
+  bool dostatus;
+  bool m_producer_pALPIDEfs_not_ok;
+  bool m_producer_pALPIDEss_not_ok;
+  bool m_startrunwhenready;
```
simonspa: Please refrain from changing indentation since this makes it nearly impossible to detect changes. We are aware of the fact that some code is really badly layouted and we are planning to run a beautification tool before releasing 1.5.

gui/src/euRun.cc
```diff
@@ -219,7 +232,7 @@ void RunControlGUI::OnConnect(const eudaq::ConnectionInfo & id) {
     SetState(ST_NONE);
   }
   if (id.GetType() == "LogCollector") {
-  	 btnLogSetStatus(true);
+    btnLogSetStatus(true);
   }
 }
```
simonspa: See commend above about changing indentation.

main/lib/CMakeLists.txt
```diff
@@ -14,7 +14,18 @@ endif (ROOT_FOUND)
...
+     FIND_PACKAGE( TINYXML REQUIRED )
+     if (TINYXML_FOUND)
+       INCLUDE_DIRECTORIES( ${TINYXML_INCLUDE_DIRS} )
+       SET(ADDITIONAL_LIBRARIES ${ADDITIONAL_LIBRARIES} ${TINYXML_LIBRARIES} )
+       ADD_DEFINITIONS(-DUSE_TINYXML)
+       message("Compiling main library using TinyXML")
+     endif (TINYXML_FOUND)
+endif (USE_TINYXML OR BUILD_palpidefs)
+
+list(REMOVE_ITEM plugins_sources plugins/GenericConverterPlugin.cc library_sources)
```
simonspa: You are removing `library_sources` from `plugins` here. Since they don't overlap, nothing should happen currently, but might lead to weird behavior in the future.

main/lib/src/Event.cc
```diff
@@ -14,7 +14,11 @@ namespace eudaq {
       "EORE",
       "HITS",
       "FAKE",
-      "SIMU"
+      "SIMU",
+      "EUDAQ2",
+      "PACKET",
```
simonspa: @RPeschke shouldn't the `FLAG_PACKET` be gone by now? Or did you only remove it from 2.0-dev?

nreader/src/EUTelNativeReader.cc
```diff
@@ -145,7 +147,14 @@ void EUTelNativeReader::readDataSource(int numEvents) {
   }
 
   if ( reader->Event().IsBORE() ) {
-    eudaq::PluginManager::Initialize(  reader->Event() );
+//    eudaq::PluginManager::Initialize(  reader->Event() );
+    
+    const eudaq::DetectorEvent & ev = reader->Event();
+    string config = ev.GetTag<std::string>("CONFIG","Config not found");
+    size_t dutPosPlace = config.find("DUTposition = ");
+    string dutPosStr = config.substr(dutPosPlace+14,3);
```
simonspa: This seems to be something hardcoded. What is this code section supposed to do?

main/lib/src/RunControl.cc
```diff
@@ -80,8 +82,12 @@ namespace eudaq {
     SendCommand("CONFIG", to_string(config));
     if (config.SetSection("RunControl")) {
       m_runsizelimit = config.Get("RunSizeLimit", 0LL);
+      m_runeventlimit = config.Get("RunEventLimit", 0LL);
+      m_nextconfigonfilelimit = config.Get("NextConfigFileOnFileLimit", false);
```
simonspa: I don't completely get what this feature is doing. Is it correct that you want to have the RunControl automatically reconfigure with another config file after stopping a run via the RunSizeLimit and restart a new one after the reconfiguration?

Quite a useful feature I guess, especially when thinking about overnight runs scanning over a set of parameters.

However, I would suggest renaming the parameter a bit to make it more clear what it's supposed to do, maybe `NextConfigFileOnRunChange`? Also, how do I specify what the range for config files is?

nreader/src/EUTelNativeReader.cc
```diff
@@ -184,7 +193,10 @@ void EUTelNativeReader::readDataSource(int numEvents) {
 				 << "Check that eudaq was compiled with LCIO and EUTELESCOPE active "<< endl;
 	throw MissingLibraryException( this, "eudaq" );
       }
+      if (lcEvent->getParameters().getIntVal("FLAG") == Event::FLAG_STATUS) {/*cout << "Skipping event " << lcEvent->getEventNumber() << ": status event" <<  endl;*/ continue;}
+      else if (lcEvent->getParameters().getIntVal("FLAG") == Event::FLAG_BROKEN) {/*cout << "Skipping event " << lcEvent->getEventNumber() << ": broken or out of sync" <<  endl;*/ continue;}
       ProcessorMgr::instance()->processEvent( lcEvent );
+//      if (eventCounter==0) streamlog_out (MESSAGE5) << "Place of telescope:\t" << lcEvent->getParameters().getFloatVal("DUTposition") << endl;
```
simonspa: Please remove commented lines, we can always get them back from git's history.

simonspa:
Dear @freidt,

thanks for removing those files. To me it looks like the files like ` producers/palpidess/scripts/config_set_generation/conf/palpidefs/brokenPixelMasks/W1-17_broken_pixels.txt ` are really device specific -  I would love to see them removed, since this repository is about the DAQ software, and not intended to serve as database for device settings and broken pixels in certain setups.

Furthermore I went through the code changes touching EUDAQ annd added some inline comments, please have a look.

Thank you very much for your patience... :)
/Simon

simonspa:
Concerning the `RunEventLimit` and `NextConfigFileOnFileLimit` settings in RunControl: where do you actually process them? I somehow missed that code snippet while browsing the changes...

main/lib/src/RunControl.cc
dreyling: FYI: Richard recently managed to use this feature (automatic stop configure start) by using a script something like...
'''
# changes DUT Mask from current value to 2
(Get-Content .\ni_autotrig.conf) -replace 'DutMask.*', 'DutMask = 2' | Set-Content .\ni_autotrig.conf -Encoding Ascii

# to change a parameter for all files in the folder:
ls *.conf |foreach{ (Get-Content $_) -replace 'DutMask.*', 'DutMask = 2' | Set-Content $_.FullName -Encoding Ascii }
'''
please refer to Richard for more details

simonspa: afaik @RPeschke uses the feature of the run control being scriptable and does not rely on above mentioned feature - but correct me if I'm wrong.

freidt: @simonspa The feature does exactly what you guessed. Once the run is changed the next config file is picked. The range is from where you start the first run with this enable in the GUI until there are no more config files left (in alphabetical order). The actual changing is done in euRun.hh.

nreader/src/EUTelNativeReader.cc
freidt: This was a temporary fix curing a short coming in our producer for the pALPIDEfs. I've removed this lines again in a new commit

gui/include/euRun.hh
freidt: @simonspa do you wan't me to revert that? Is it fine, if I provide a diff without any whitespace changes?

main/lib/src/Event.cc
RPeschke: this flag is the replacement for the packet class in eudaq 2.0. therefore it has to stay in

main/lib/CMakeLists.txt
freidt: @simonspa I've reverted that part of the change.

gui/include/euRun.hh
freidt: For the moment, in order to allow quick look, I've uploaded a diff containing the output of the following
```
git diff -w  upstream/v1.5-dev  -- gui/
```
to
https://cernbox.cern.ch/public.php?service=files&t=19062b6e4403db0032ea36085ec2768e

freidt:
Hello @simonspa ,

- The config_file_generation folder has completely been moved from the pALPIDEFs producer.
- `NextConfigFileOnFileLimit` has been renamed to `NextConfigFileOnRunChange`
- `RunEventLimit` and `NextConfigFileOnRunChange` are processed in `euRun.hh`.

Cheers,
Felix

gui/include/euRun.hh
simonspa: Thanks for the diff, a few comments on that:

```
@@ -74,13 +73,9 @@ class RunControlGUI : public QMainWindow, public Ui::wndRun, public eudaq::RunCo
 
   void SetStateSlot(int state) {
     btnConfig->setEnabled(state != ST_RUNNING);
-			btnTerminate->setEnabled(state != ST_RUNNING);
     btnStart->setEnabled(state == ST_READY);
     btnStop->setEnabled(state == ST_RUNNING);
   }
-
-  void on_btnTerminate_clicked() { close(); }
-
   void on_btnConfig_clicked() {
```
Does this solve a problem you experienced? What is the reason behind these changes?

```
@@ -151,6 +163,10 @@ signals:
   status_t m_status;
   int m_prevtrigs;
   double m_prevtime, m_runstarttime;
-    int64_t m_filebytes;
+  long long m_filebytes;
+  long long m_events;
```
We try to stick to fixed size integers to keep maximum compatibility. Is there a specific reason for changing to `long long` or would `uint64_t` also be fine with you?

```
+  bool m_producer_pALPIDEfs_not_ok;
+  bool m_producer_pALPIDEss_not_ok;
[...]
+    else if (id.GetName() == "pALPIDEfs") {
+      m_producer_pALPIDEfs_not_ok = (status->GetLevel() != 1);
+    }
+    else if (id.GetName() == "pALPIDEss") {
+      m_producer_pALPIDEss_not_ok = (status->GetLevel() != 1);
+    }
+  }
```
Is this really necessary? I know that the current finite state machine of the RunControl is somehow broken, but we will have a Summer Student working on this. I hope that after that we will always be in a well-defined state and the producers returning error messages will not just be a nice GUI plotting feature.

main/lib/src/Event.cc
simonspa: Okay, fine then.

simonspa:
Cool, we're getting there! :)

gui/include/euRun.hh
freidt: The first change was unwanted and due to mistake while rebasing/merging. (fix committed)
The second as well. (fix committed)
The third is unfortunately necessary in order to allow the automatic config changes. Our producers need some time (up to minutes) for configuration. Having them in a non defined state when a run is started lead to crashes. That's why we implemented the safety mechanism here. Would be fine with you to remove this as soon the new FSM is in place? It's anyways written in way which should not affect other producers.

simonspa: Understanding what the reason is, I'm fine with including it. Maybe you could add a comment to the respective lines describing the problem and noting that it can be removed once the FSM is properly working.

freidt: @simonspa: I've added a comment to the member variable declaration and repeated it in the commit message.

simonspa:
Okay I guess we are almost ready to go.

I don't really have an opinion about the STARTRUN and KILLRUN scripts in `/etc/scripts`. If @RPeschke does not object I would probably leave them in.

I went through all changes once more, and all looks fine to me now. Only maybe the files in
https://github.com/freidt/eudaq/tree/master/producers/palpidess/scripts/config_set_generation/conf/palpidefs/trackingConf
could be removed, but it's not that important.

After hearing from @RPeschke I would merge this PR.

freidt:
Hello @simonspa and @RPeschke,
I've just added two small changes to our converter plugin.
Cheers,
Felix

freidt:
Hello @simonspa and @RPeschke,
I've just added a couple of bug fixes to our producers. Can you already estimate, when you will merge this pull request? Is there anything additional we can provide to ease the review?
Thanks a lot in advance!
Cheers,
Felix

simonspa:
Merged, thanks @freidt and all other authors.

PR #136, opened 2015-05-23
Fix wrong message in case of missalignment.

adrianf0:


simonspa:
Dear @adrianf0 

thanks for your patch, could you please file the PR against the branch `v1.5-dev` instead of `v1.4-dev`?

Thanks,
Simon

adrianf0:
Dear Simon,
The last official tag is 1.4.5. The Desy telescope we are using in H6 at CERN uses this version of software. I commit patches which came up during the telescope usage. I thought they should go to the v1.4-dev and later they will be promoted to the next release. Am I wrong ?

Regards,
Adrian

simonspa:
Dear @adrianf0 

you are both right and wrong. :)

Indeed v1.4-dev is the stable branch for the v1.4 series of EUDAQ. But since quite some changes piled up lately which are not compatible with calling them a "bugfix point release" we decided that the next EUDAQ release will bear the name "v1.5", hence the new branch.

With v1.5 released, v1.4 will be discontinued.

Cheers,
Simon

adrianf0:
Are you planning merge/cherry-pick of some commits from v1.4-dev ? For me it's straightforward to request pulls for commits which have been tested on production system. Otherwise I would need to test them on other branches (v1.5, v2.0, master) and I don't really have such possibility (the production system uses 1.4.5). I don't understand work flow you want to follow. I understand that master and v2.0 follow a newer architecture of the DAQ system (data driven producers). Where should I commit CLICpix and Timpix3 specific patches which I would like to apply also for newer releases of the EUDAQ software ? Where should I commit bugs spotted during usage of the latest official release ?

simonspa:
Dear @adrianf0 

you are actually hitting a weak point here. We are currently in consolidation to bring the structure of the repository back to a sane state where stable development is possible and comprehensible. We ended up in this situation because at some point in the past the development was not very nicely coordinated and people started merging code into various branches where it didn't belong.

The workflow we are aiming at would be:
* having one branch for the currently used EUDAQ version with ongoing development: `master`
* having one branch for every stable release, like `v1.4-dev`
* branching off from master whenever a new stable version is released: `v1.5-dev` etc
* having one (or several) branches for future, data driven versions of EUDAQ : `v2.0-dev` or similar, which will be merged into `master` as soon as it's time for it. Of course this will be kept up to date with the latest `master`
* all patches and bugfixes to stable branches are cherry picked into `master`

Currently this is obviously not the case. Branch `v1.5-dev` is based in `v1.4-dev` instead of master and contains all fixes that ever went into this stable branch. After cleaning up with `master` everything will be a bit more clear I hope.

Which means: currently I try to get everything into `v1.5-dev`, which should be no problem since it's currently still almost the same as `v1.4-dev` just with some updated producers and a few bug fixes.

So, as long as you're not making explicit use of data driven methods from EUDAQ2, all Timepix and CLICLpix updates should go into `v1.5-dev` which will eventually become the next stable version of EUDAQ1.x.

I hope I managed to explain a bit our situation... :)
/Simon

adrianf0:
Dear Simon,
Thank you for you explanation. Now it is more clear. Last question: my last commits, when v1.4-dev becomes a stable branch, will be moved by the main developers to v1.5-dev and will become part of the next release ? Or you expect I will cherry pick them by myself first to the master branch and after merge the master with v1.5-dev ?

Regards,
Adrian

simonspa:
You won't have to do anything, we will carefully cherry-pick everything.

PR #141, opened 2015-06-22
Bug in the pALPIDEfs converter

mkofarag:
Hi,

We found and corrected a bug in our converter for the pALPIDEfs chip and we would like to merge the change to the official repository.

Best regards,
Monika


simonspa:
Hi @mkofarag,

thanks for the update. Merged.

/Simon

mkofarag:
Thank you for the very quick reply!
Best regards, Monika

PR #142, opened 2015-06-22
Bug fixes for converter, loser timestamp cut

markuskeil:
Local changes in pALPIDEfs producer and converter 

Thanks in advance for merging


simonspa:
Sure, no problem!

Issue #76, opened 2014-06-19
Online Monitor Crashes during run


RPeschke:
the gui thread of the Online monitor Crashes directly after the worker thread has called this function  calling the first time this function:

void OnlineMonWindow::registerTreeItem(std::string item) 

call stack:
OnlineMon.exe.exe!OnlineMonWindow::registerTreeItem(std::basic_string<char,std::char_traits<char>,std::allocator<char> > item) 
 	OnlineMon.exe.exe!HitmapCollection::registerPlane(const SimpleStandardPlane & p) 
 	OnlineMon.exe.exe!HitmapCollection::fillHistograms(const SimpleStandardPlane & simpPlane) 
 	OnlineMon.exe.exe!HitmapCollection::Fill(const SimpleStandardEvent & simpev) 
 	OnlineMon.exe.exe!RootMonitor::OnEvent(const eudaq::StandardEvent & ev)
 	EUDAQ.dll!eudaq::Monitor::ProcessEvent()
 	EUDAQ.dll!eudaq::Monitor::OnIdle()

Issue #17, opened 2013-12-19
OnlineMonitor not working/tested on Windows


hperrey:
The OnlineMonitor has not yet been successfully been built/tested on Windows.

Need to accomplish:
- [x] Compile OnlineMon on Windows using CMake
- [ ] Document installation procedure (i.e. ROOT) in README and manual
- [ ] Test OnlineMonitor on NI crate in testbeam/lab

arummler:
After the latest fix the OnlineMonitor compiles again but it crashes with the same sympton as described in this issue. I compiled it using Visual C++ 2013, Qt 5.5 and root_v5.34.32.win32.vc12.debug and used the debug configuration. The crash appears within ROOT after the "theApp.Run()" has been called in file "TGWin32.cxx", method "TGWin32::GetWindowSize", line "if (GDK_DRAWABLE_TYPE(id) == GDK_DRAWABLE_PIXMAP) {". Does anybody have an idea what's happening? Otherwise, I will try to put together a minimal example and post a bug report on the ROOT tracker.

![runtime_err1](https://cloud.githubusercontent.com/assets/2855671/8847099/0aaaf70a-312c-11e5-8d90-6dbfe2d975c2.png)
![runtime_err2](https://cloud.githubusercontent.com/assets/2855671/8847103/0ab24c1c-312c-11e5-8d42-9d6c13dfce17.PNG)
![runtime_err3](https://cloud.githubusercontent.com/assets/2855671/8847100/0aad8e84-312c-11e5-9765-bf63caad200a.PNG)
![runtime_err4](https://cloud.githubusercontent.com/assets/2855671/8847102/0aaee11c-312c-11e5-8dc8-15d264771dfc.PNG)
![runtime_err5](https://cloud.githubusercontent.com/assets/2855671/8847101/0aada374-312c-11e5-9a8d-7a3181c2e411.PNG)

arummler:
Okay, it seems it does no longer occur if the lines
Cvs_left->MapSubwindows();
Cvs_left->MapWindow();
in void OnlineMonWindow::registerTreeItem(std::string item) are taken out. But the tree is not shown as well. Furthermore, there occur sporadically other run time errors when it is attempted to access elements beyond a vector. Solved this by replacing reserve by resize and rewriting the getWidthX and getWidthY methods of SimpleCluster.
[...]

PR #144, opened 2015-07-04
Applied changes made by Martin Kocian for the FE-I3 readout with the …

arummler:
…RCE (+HSIO+MCC). This also fixes the missing ToT and LVL1 histograms in the OnlineMonitor for this kind for data aquisition. Started testing in testbeam July 3rd at CERN with ATLAS telescope, needs more testing (up to now only existing FE-I4 functionality used) and potentially cleanup.


simonspa:
Not merging this due to the "needs more testing and cleanup" - I'm expecting more commits from your side. If this is a misinterpretation and should be merged, let me know. It compiles - and since it's a user processor we don't care about much more. :)

arummler:
I think the processor commit can be merged as there were no problems during data taking. If anything occures during the ongoing data analysis we can still revert or fix. I tried to run the OnlineMonitor on Windows 7 and it did not compile. Introduced two small fixes which helped. Please trigger a Travis rebuild manually as it failed to download during the first attempt, a svn hickup, and I don't want to push a dummy commit or close and reopen the PR. Now the OnlineMonitor crashes irregularly after startup (running in read from file mode) but I will open an issue for that.

eudaq_compilation_error_1
eudaq_compilation_error_2

arummler:
The Windows thing is follow-up of issues #17 and #76.

monitors/onlinemon/include/SimpleStandardEvent.hh
```diff
@@ -11,7 +11,7 @@ typedef char int8_t;
 #include <string>
 #include <vector>
 #include <iostream>
-#include <stdint.h>
+//#include <stdint.h>
```
simonspa: Is there a reason for this change?

arummler: It removes the nine errors in the second screen shot and it seems it is not necessary as it compiles without including stdint.h. The include files that appear in the error message are (indirectly) included by stdint.h.

simonspa:
@arummler okay, is fine then and also compiles. TravisCI failed due to SourceForge's downtime last week and the ROOT package we pull from there.

I'm sorry that we just released eudaq 1.5.0 without the PR, but I'll merge now and it will be part of EUDAQ v1.5.1 soonish I guess. Maybe the WIndows issue is also solved until then.

PR #150, opened 2015-07-21
V1.5 dev merge with sct and root interface

RPeschke:
Changes from the atlas strip testbeam in May. Espeecially the changes to the ROOT_interface are necessary because they allow to store information at the end of run event which was not possible before. 

In addition to this there are some changes to the SCT_DATA_convert which allow the conversion to LCIO


simonspa:
Looks fine to me, merging.

PR #165, opened 2015-12-01
Master

airqui:
AHCAL version to pull to the EUDAQ main repository.
Calice Producer is written for ScCal (a draft for SiCal)
CaliceGenericConverterPlugin is written for online/offline conversion of raw data to a size-optimized SLCIO format.


simonspa:
Hi @airqui 

thanks for your pull request, I browsed through the changes and have a few comments which I would like to see addressed.

* You add a new version of the FileWriter (`FileWriterLCIOC.cc`) which I compared to the standard version. Beside some indentation changes only a few lines of code seem to be changed. Wouldn't it make more sense to add these changes to the original converter in order not to add unnecessary code?
* As discussed, your new producer folder "calice" can contain whatever you want - however, this "whatever" does not extend to file types generally banned from the repository (binary files, object files) as well as to obviously useless files (log files, stack traces). Please check/remove the folder `runscripts` which contains lots of code, compiled binaries as well as other seemingly useless files such as `ScReader.backup` or `crashlog.txt`.
* Probably it would be better to reset and circumvent commit ff1c454 - otherwise all these binary files will stay in the git history and thus in the repository forever.

Let me know if you have questions or disagree.
Cheers,
Simon

notify: @dreyling @RPeschke

airqui:
Dear @simonspa 

Thanks for your reply, and sorry for my late answer. I decided to keep a separate converter FileWriterLCIOC to not conflict with the FileWriterLCIO. First because we want to have a specific data name output and second because we have a different treatment of the LCIO headers for our LCIOGenericObjects. I think that both converters are not compatible.

About the runscript folder you are completely right. Probably the best option is to reset the ff1c454 commit as you say and commit it again deleting all these files... how can I do this without destroying everything?

Cheers,
Adrian

simonspa:
```
# brings you back to before the messy commit
git checkout 5da5589
# start new branch
git checkout -b ahcal_contribution
# newly add the important files without the logs etc:
git add file1.cc file2.cc file3.h file4.h
git commit -m "Add AHCAL code"
# Add the last commit with the converter plugins again:
git cherry-pick c38d470
# publish the new branch
git push origin ahcal_contribution
```

and you should be fine. Then just open a new pull request for this newly created branch

simonspa:
Closing, superseded by #170

PR #169, opened 2015-12-15
Fix compilation error with ROOT6. The histogram bit TH1::kCanRebin ha…

Eyiliu:
Fix compilation error with ROOT6. The histogram bit TH1::kCanRebin has been removed in ROOT6.


simonspa:
Thanks!

PR #170, opened 2016-01-12
Ahcal contribution

airqui:
cleanup of the AHCAL code with a dummy version of the GetStandardSubEvent class.


RPeschke:
Hello,

Thanks for you contribution. There are a few thinks I would like to discuss before we can merge the pull request. 



1. main/include/eudaq/CaliceGenericConverterPlugin.hh 
we usually don't have header files for converter plugins. The idea is that everything is contained in one single file. Do you need the header file somewhere? if not you can just copy the code to the .cc file. 



2. producers/calice/include/CaliceProducer.hh  uses Pthread.

Pthread is not standard therefore I would rather have it replaced with std::threads and std:: mutex 



3. producers/calice/runscripts/CaliceDataCollector.cc 

I don't quite see for what you need to have your own data collector. If there is a feature missing in the data collector we should discuss how we get it to the main library. 



4. producers/calice/runscripts/CaliceGenericConverterPlugin.cc 
-do you have two converter plugins? it seems there is one in producer and one in plugin folder. Again the converter plugins do not have header files.

airqui:
On 01/13/2016 02:25 PM, RPeschke wrote:
>
> Hello,
>
> Thanks for you contribution. There are a few thinks I would like to 
> discuss before we can merge the pull request.
>
> 1.
>
>     main/include/eudaq/CaliceGenericConverterPlugin.hh
>     we usually don't have header files for converter plugins. The idea
>     is that everything is contained in one single file. Do you need
>     the header file somewhere? if not you can just copy the code to
>     the .cc file.
>
> 2.
>
>     producers/calice/include/CaliceProducer.hh uses Pthread.
>
> Pthread is not standard therefore I would rather have it replaced with 
> std::threads and std:: mutex
>
>  1. producers/calice/runscripts/CaliceDataCollector.cc
>
> I don't quite see for what you need to have your own data collector. 
> If there is a feature missing in the data collector we should discuss 
> how we get it to the main library.
>
>  1. producers/calice/runscripts/CaliceGenericConverterPlugin.cc -do
>     you have two converter plugins? it seems there is one in producer
>     and one in plugin folder. Again the converter plugins do not have
>     header files.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/eudaq/eudaq/pull/170#issuecomment-171288715>.
>
Hello Richard,
Thanks a lot for your comments. I will remove the header as you say and 
replace the Pthread as you suggested.

About your last comment, you are completely right, but this is for the 
"eudaq_calice_standalone" code that we have been using during 2015 
testbeam. The new version is much more optimized and properly integrated 
into the eudaq framework (i.e. using the eudaq standard collector).
I don't compile this but I put it in the runscript folder for 
completeness and also to be able to use it when needed so we keep a 
kind. I know that this is not the most ellegant way to proceed and we 
have another private git repository with these files, but I think that 
is good to have everything in the same repository. I will try to make it 
clearer and put it in a separated folder.

Cheers,
Adrian

airqui:
Hello

I made all the changes that you proposed.
Can you take a quick look to the CaliceProducer parts where I have replaced the pthread by thread and mutex? It works but I am really not an expert in this... 

Thanks a lot,
Adrian

RPeschke:
Hi,


1. the member function of a mutex are never called directly. they are always called through a RAII class
use  lock_guards: http://en.cppreference.com/w/cpp/thread/lock_guard


2.  I don't understand construct like this:
void MainLoop_t(){  
   std::thread mainloopthread (&CaliceProducer::MainLoop,this);  
   mainloopthread.join();  
 }  
why is it on a different thread?


3. absorb port handles in RAII classes
 int ret = connect(_fd, (struct sockaddr *) &dstAddr, sizeof(dstAddr)); 
_fd is a handle which can leak if you dont protect it with RAII

airqui:
Hi Richard,
I think that now we understood a bit better how to use the RAII classes. I hope that now is okay.
Cheers,
Adrian

RPeschke:
Hi Adrian,

yes this is better but there are still a lot of very long function. can you reduce the max function size to 20 lines and not more than one loop. 


there are still "new" commands in your code can you replace them with make_shared or make _unique 

- About your last comment, you are completely right, but this is for the
"eudaq_calice_standalone" code that we have been using during 2015
testbeam. The new version is much more optimized and properly integrated
into the eudaq framework (i.e. using the eudaq standard collector).
I don't compile this but I put it in the runscript folder for
completeness and also to be able to use it when needed so we keep a
kind. I know that this is not the most ellegant way to proceed and we
have another private git repository with these files, but I think that
is good to have everything in the same repository. I will try to make i clearer and put it in a separated folder.

I am not sure that this has to go in the main repository. it looks like something very specific to you plan to reuse it? but any way things that should go in the main repository are either things that have a benefit for all users or the converter plugins but this is a designe mistake in the library.  For me this looks to specific to go to the main repo.



Cheers,
Richard

airqui:
Hi Richard,

Let's see how it looks now. I have "cleaned" a lot of long functions and removed the "Standalone" folder. For the SiReader.cc I did not change a word since it is not from our scintillator detector but from other group (Silicon based). When we have this code in the official eudaq, I will ask them to check it and optimize it in the direction that you point.

About this:
"there are still "new" commands in your code can you replace them with make_shared or make _unique "
Can you develop a bit more what you mean with that sentence? 

Thanks a lot!
Adrian

simonspa:
Hi @airqui,
thanks for all the work and cleanup, I would now be fine with merging - additional improvements can still go in later any time.
@RPeschke do you have any objections?

Cheers,
Simon

RPeschke:
Hi,

Yes it looks much better now. 
The only thing left to change is that this function needs to be shorter:

virtual bool GetLCIOSubEvent(lcio::LCEvent &result, eudaq::Event const &source) const


Also I would like to make the style more uniform but maybe we do it with our scripts after the merge. @simonspa what do you think.

airqui:
Hi @RPeschke, @simonspa 

You are right, this function is long. This is due to the "special" format of the temperature data. This should be uniformized and changed it to be compatible with the "data" collection and this is being done but it needs the agreement of all the subgroups and need to be done coherently in all the other codes (analysis, reconstruction, etc...) and moreover, we need to keep as much as backward compatibility as possible with previous testbeam files, so:
 - Yers, we are changing the LCIO format of all collections and therefore the virtual bool GetLCIOSubEvent(lcio::LCEvent &result, eudaq::Event const &source) const
- but we are not doing it now since it will take a while to get an agreement about the final lcio format.

Moreover, we are working with the Sussex people to create kind of "calice-standardplanes" objects. Using these objects, the length of the functions inside the plugin will be also reduced. So, I would really appreciate if you take this plugin as it is at the moment so all people from the collaboration can contribute.

Cheers,
Adrian

simonspa:
So let's settle with this, further improvements to come later. Thanks, @airqui!

simonspa:
I'm closing this PR since it was filed against the `v1.5-dev` branch. Since yesterday we have the masterbranch up to date for EUDAQ 2.0, so I manually merged it to there.

PR #171, opened 2016-01-18
Issue 168

arummler:
Exchanged file list by standard file dialogue solving issue 168. Implemented saving of program settings.


simonspa:
I was about to say "very cool" but after compiling and testing I would like to correct: "awesome".
Thanks a bunch for looking into this!

PR #172, opened 2016-01-18
Automatic sensor ID assignment introduced in base class and added to …

arummler:
…APIX-CT and USBPIXI4 data converters. Furthermore, the USBPIXI4 data converter now handles several instances of this producer. Due to lacking data in the data stream it needs to make the assumption that the data are always recorded in the same order. This seems to work but will be make easier by extending the recorded information for future data taking.
Missing:
- USBPIXI4 needs to adapted to changes in producer (version number, producer number tags) as soon as those are done
- APIX-CT and USBPIXI4 should have a possibility to assign deliberately sensorIDs (connected to changes in producer)
- USBPIX needs the same changes
- Make other data converters use the sensorID assignment mechanism as well.
Option:
It is not so nice that the acquired sensor ids have to be released explicitly by the data converter. This is due to the fact that the same instance of a converter is used for consecutive runs. In practice it does not make a difference as usually the number and type of sensors stay the same for consecutive runs which are handled at the same time.


main/lib/plugins/DataConverterPlugin.cc
```diff
@@ -60,4 +63,79 @@ namespace eudaq {
...
+			std::cout << "[EUDAQ:ConverterPlugin:BaseClass] SensorID " << sensorCounter << " has been assigend." << std::endl;
+			#endif	
+			return sensorCounter;
+		}
+		sensorCounter++;
+	}
+	#if USE_LCIO && USE_EUTELESCOPE
+	streamlog_out(ERROR9) << "No sensor ID assigned - this should not be possible but it occurred!" << std::endl;
+	#else
+	std::cout <<  "No sensor ID assigned - this should not be possible but it occurred!" << std::endl;
```
simonspa: :)

simonspa:
@arummler finally I got around reading your code - sorry for the delay.
I think I need a bit more of explanatory comments here. I see the sensor assignment routine in the base class - but how do I for example make sure to have the different parameter sets from the various BORE events available for the different detector instances? I see that you implemented this in the USBPIX converter but it looks quite cumbersome to me.

Wouldn't it finally be better to try to get rid of the singleton instances for convertes? That way the data streams would be nicely separated in different class objects and we wouldn't have to juggle with vectors of BORE configurations.

I know we discussed this at the last workshop and couldn't come up with a convenient way to rework the plugin factory for converters, but I still believe that would be the better way to go.

However, having this code in as an intermediate solution wouldn't hurt either I guess...
Opinions?

/Simon

arummler:
@simonspa I tried to make a minimal change, thus the assignment of Sensor IDs gets only managed between converter plugin instances. Everything else the producers have to solve themselves (including shifting through the BOREs). I agree that the solution in the plugins which I edited is particularly ugly due to some missing tags (hopefully to be improved for next time). We can try to make the move away from the singleton solution and see whether it works out. For the time being I would be happy if my change would go in because our users need it to process some of last year's data. In the moment they have to use a fork from my github. There is also the issue of the logging output which could be also solved more elegantly as we discussed last year. But for that I can open a new issue.

simonspa:
Hm, I'm not particularly happy with this solution, also in the hindsight of a v1.6 which might be released not so far in the future.

Do we have any other opinions? @RPeschke ?

RPeschke:
I have not yet fully understood everything but I am not happy with introducing another function to the data converter ...

what exactly do you need. if it is just a unique id for that specific instance of the plugin eudaq 2 has actually a virtual function in the data converter that can do this. 

I am also not happy about streamlog since it is another dependency...  

is there any change with wich you can achieve your goals without having to change the core library?

Cheers,
Richard

simonspa:
Well, the streamlog dependency is only present when using EUTelescope - which depends on streamlog anyway.

arummler:
streamlog: @simonspa is right, I only use streamlog in the converters if they are linked against Eutelescope and otherwise just std::cout. This is to avoid the "\[unkown\] (sic!)" output. The duplication of output lines is certainly not optimal and I think your proposal during the EUDAQ workshop was to have some C++ logging library in EUDAQ which enables just to switch between those two outputs.

> what exactly do you need. if it is just a unique id for that specific instance of the plugin eudaq 2 has actually a virtual function in the data converter that can do this.
>
> I am also not happy about streamlog since it is another dependency...
>
>is there any change with wich you can achieve your goals without having to change the core library?

I will explain the general situation. Within ITK we are using (beyond the NI and TLU) three/four different producers and subsequently converter plugins: APIX-CT, USBPixI4, USBpix and PyBAR. The first one is used for the RCE DAQ hardware and the last three ones for the USBpix DAQ hardware. Why three different converter plugins for the same hardware? USBPixI4 and USBpix handle two different frontend chips: FE-I4 and FE-I3. Until recently they were serviced by two different forks of the same software - STcontrol. In the latest version they were unified in one but there are still two different converter plugins. Then there is a alternative, low weight, DAQ software for USBpix - PyBAR.

Last year we had two testbeam periods where we were running more complex setups. Particularly, two basic scenarios (with different mixtures):
1. RCE and USBPix --> APIX-CT and USBPixI4
2. Multiple instances of the USBpixI4 producer
Now the analysis posed some problems due to mixing of different sensor planes. The USBpixI4 converter plugin is not equipped to deal with several instances of the corresponding producer. This looks a bit more severe because the raw events are not properly tagged and one has to rely on the order of the events being written into the data stream. This seems to work now and for next time we want to extend the tagging in the producers in order to avoid this problem.

But then there is still the same problem as with producers of different types: The assignment of the SensorID is the responsibility of the converter plugins and in our case they are using information assigned during data taking by their corresponding producers. Unfortunately, the different producer instances do not know about each other and the user has to set the SensorIDs basically correctly by hand in the config file. This was not done correctly. 

Now, for my take on the problem: I introduced a method in the converter plugin base class which provides on request an unused SensorID and remembers used ones until they are freed by the convert plugin. I was not aware that in eudaq 2.0 there is now the method getUniqueIdentifier which was mentioned by @RPeschke. This method does something similar but is limited to assign a range of 10 SensorIDs to a producer plugin. The producer plugin is reposible to distribute the SensorIDs within this range and to keep track. I could use that instead but: It is part of eudaq 2.0, in the moment the users that need to analyze their last year's data are on my fork. If I adapt my converter plugin changes to use this method shall they switch to eudaq 2.0? Is it production ready? And shall we keep the singleton pattern?

arummler:
If we can backport getUniqueIdentifier to 1.5 I will rewrite it accordingly.

RPeschke:
>This method does something similar but is limited to assign a range of 10 SensorIDs to a producer plugin. 

this can easily be changed. we can make it a 100 or 1000 or more. 

> It is part of eudaq 2.0, in the moment the users that need to analyze their last year's data are on my fork. If I adapt my converter plugin changes to use this method shall they switch to eudaq 2.0? Is it production ready?

I would say it is in beta stage. I would not switch to it right before a test beam but for the analysis I think it is safe to try. 

For test beam use it really depends on how good you understand your system. If you have still bugs in your system I would not switch. But if you understand your system so well that you can be sure that all errors have to come from eudaq then you can switch. Of course I would like to see it used more often. 

> If we can backport getUniqueIdentifier to 1.5 I will rewrite it accordingly.

This can be done but I would do it after we have tried eudaq 2. 

> And shall we keep the singleton pattern?

Yes. Otherwise we would break to much legacy code.

arummler:
@RPeschke Okay, then I will change the converter plugin on master. Whether to backport it to v1.5 we can see later on. Regarding the use of 2.0 during the data taking: We (ITK) have this year a bit of time when we could try out stuff at the SPS while being parasitic without loosing testbeam time. In the moment it is foreseen to test new DAQ systems (HSIO-II and MMC3) but I suppose we could also do a couple of test runs with the master branch. What would be most interesting (and enticing to make the switch) is naturally to go for the big feature of 2.0 - the separate readout frames of telescope and FE-I4s. But to my understanding we have to wait for the right TLU firmware first.

RPeschke:
There are two things to test
- first does it work at all. do we have any crashes etc. but also does the logic does anything wrong do we miss events or so.

- the second point is testing the merging of data streams. you are right with one of the current TLUs we cannot do it but with two TLUs we can. 
we can have two TLUs running cascaded. we did this already once at desy but not very systematic and also only one run. I would be really interested in what we would get at CERN. 
 

folie32
folie33
folie34

arummler:
- Sure, this is no problem.
- I would be interested in trying that. The schematics I remember from your EUDAQ workshop talk. I think I might be able to get a second TLU (either AIDA telescope H6B if not used or Paris) but could you give me more information about how to plug it together and how to configure it? I never used trigger out on the current TLU - is it just one of the RJ45 connectors reconfigured? If you tried it with FE-I4, do you have already an adapted producer?
img_3498

RPeschke:
actually it is pretty simple you just take the lemo output from one TLU and plug it into the next TLU. There is a configuration key to take lemo. I have vergotten the key to tell the TLU to not require handshake. but you can just feed the trigger output to busy input from the same TLU. 

when I did the test I was running two sets off eudaq completely independent. with different data collector and run control. you then start the TLU eudaq wait untile everything is setup and then you start the main eudaq. that way both TLUs have the same first event which is important to synchronize it later. 
only at the converter stage you merge the two files. 

 



[Producer.TLU]
OrMask = 0
VetoMask = 0
...
DUTInput3=LEMO  <---
...

arummler:
Closing it for the time being, will open a new PR with changes using getuniqueidentifier.

PR #175, opened 2016-01-20
clean up of the  Converter

RPeschke:
I overlooked some debug code that is removed with this commit 
& 
Clean up


simonspa:
Plus adding `void add_merger()`?

RPeschke:
This is part of the clean up. it just absorbs the if statement

simonspa:
Okay, after actually reading the code it makes sense... :)

PR #179, opened 2016-01-27
Converter plugins for generic producer and ccplf producers

theresaobermann:
Converter plugins for generic and ccplf


simonspa:
Everything fine, merged and ported to the master branch, too.

PR #180, opened 2016-01-27
fix some compile problem

Eyiliu:


simonspa:
Hi @Eyiliu,

do eed4cb0a241565736c33f3f3d7dfa3cfd00b6b2e and 1c54a962797955cfc86032c1db6d013b9771ac4c also apply for v1.5 of EUDAQ?

@RPeschke  	e11b0cd is fine?

RPeschke:
yes you are right. i have no clue how i can have overlook this

PR #193, opened 2016-02-20
Master: C++11 recognition realized by cmake, osx build now works, fixed bunch of warnings on MSVC

arummler:
Master: With cmake 3.1 now the C++11 flags are set automatically. OSX suddenly works again. (minus AIDA stuff that no longer exists on 2.0)


arummler:
Okay, it seems that in 2.0 there are some heavy changes in FileWriter that break OSX builds --> this PR needs more work.

arummler:
Seems that aux_source_directory is used in a non-intended way:
https://cmake.org/cmake/help/v3.0/command/aux_source_directory.html

arummler:
I am not sure why the templates work differently with gcc and clang but there is a bunch of dicussions on stackoverflow, e.g. http://stackoverflow.com/questions/1022623/c-shared-library-with-templates-undefined-symbols-error

PR #196, opened 2016-02-24
Fixed MSVC warnings

arummler:
There are still about 28 left. More commits to come.

### DIFF
--- a/main/exe/src/TestRunControl.cxx
+++ b/main/exe/src/TestRunControl.cxx
@@ -57,8 +57,8 @@ void * worker_automatic_run(void * arg) {
 
 
         rc->StartRun("");
-        std::cout << "sleep for " << (unsigned int) duration << std::endl;
-        eudaq::mSleep((unsigned int)duration);
+        std::cout << "sleep for " << static_cast<unsigned int> (duration) << std::endl;
+        eudaq::mSleep(static_cast<unsigned int> (duration));
 
         rc->StopRun();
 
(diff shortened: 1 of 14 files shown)


simonspa:
okay, waiting with merge then. Thanks for the work!

arummler:
@simonspa You are welcome. Now down to the 11 in the converter plugins (the bit shift operations, arghh) which are mentioned in the issues - please pull. I will add the changes as well to 1.5 and open a new PR.

main/exe/src/TestRunControl.cxx
```diff
@@ -57,8 +57,8 @@ void * worker_automatic_run(void * arg) {
 
 
         rc->StartRun("");
-        std::cout << "sleep for " << duration << std::endl;
-        eudaq::mSleep(duration);
+        std::cout << "sleep for " << (unsigned int) duration << std::endl;
+        eudaq::mSleep((unsigned int)duration);
```
simonspa: don't we want proper casting via `static_cast<>()`?

arummler:
You are right - I was using the sloppy old C style casting. But I think I did it at other places as well, let me check before merging.

arummler:
Done

arummler:
Appveyor build fails with code 1 --> When I connect with rdesktop to the virtual machine it works. There are some older thread in the help board which look similar, probably I will open also one.

arummler:
Need to check first why it is failing.

Issue #197, opened 2016-02-24
CaliceGenericConverterPlugin


arummler:
This converter plugin produces a warning with MSVC.
`C:\projects\eudaq\main\lib\plugins\CaliceGenericConverterPlugin.cc(71): warning C4018: '<' : signed/unsigned mismatch [C:\projects\eudaq\build\main\lib\EUDAQ.vcxproj]`

It should be enough to change line 67 from
`int nblock =4; // the first 4 blocks contain information`
to
`unsigned int nblock =4; // the first 4 blocks contain information`
Could the owner of the plugin please check that?

simonspa:
@airqui could you please check this?

airqui:
no problem, il do it.

arummler:
Checked that this is fixed for v1.6-dev.

PR #201, opened 2016-02-25
Solve warning with MSVC in CaliceGenericConverterPlugin 

airqui:
following @arummler  advice #197


arummler:
Hello, thanks for the fix. I think the appveyor failure was only a temporary timeout on the mentioned URL. It seems to work now. Probably @simonspa or @RPeschke could restart the build.

airqui:
Hi @arummler. I don't know how to pull this changes in the master branch since it is out of date in my fork...

arummler:
@airqui Hi. You can do a PR also if your branch is lagging behind. If you want to check first whether there is no conflict, do a git fetch origin (or however upstream is called in your working copy) and then git pull origin master. If there are no conflicts you should be fast-forwarded, otherwise you have to solve the merge conflicts. Probably you want to create first another branch and change to that in case it gets messed up.

PR #207, opened 2016-03-22
General Inferface of ROOTProducer with RawdataEvent

Eyiliu:


arummler:
Commit #8847c33 does not build and there is no PR for v1.5-dev branch.

main/include/eudaq/RawDataEvent.hh
```diff
@@ -76,13 +80,21 @@ namespace eudaq {
     size_t NumBlocks() const { return m_blocks.size(); }
 
     virtual void Print(std::ostream &) const;
-    virtual void Print(std::ostream & ,size_t offset) const;
```
arummler: Removal causes compilation error.

Eyiliu:
Thanks. I was not  aware the diff between PR and v1.5-dev. Fixed now.

arummler:
Seems that there is a misunderstanding as I do not see any additional pull request (PR): I just tried to keep track whether there is a corresponding PR to be applied to one branch (v1.5-dev) as well as to the other (master). Both have to be kept synchronous for the time being and Simon mentioned it is easier for him if we send a PR for both of them (otherwise he has to do it himself).

PR #239, opened 2016-06-03
IPHC RC DAQ / M26 Producer

simonspa:
*do not merge*, only meant for discussions yet.


cmake/FindThreads.cmake
```diff
@@ -0,0 +1,250 @@
+# - This module determines the thread library of the system.
```
simonspa: Why do you need this file?
We do not use pthreads since we rely on C++11 which brings its own threading.

producers/mimosa26_sbg/include/M26Producer.hh
```diff
@@ -0,0 +1,90 @@
...
+
+// EUDAQ includes
+#include "eudaq/Producer.hh"
+#include "eudaq/Timer.hh"
+#include "eudaq/Configuration.hh"
+//#include "eudaq/Utils.hh"
+//#include "eudaq/Logger.hh"
+
+// IRC DLL header
+#include "iphc_run_ctrl_exp.h"
```
simonspa: Not to be confused with https://en.wikipedia.org/wiki/Internet_Relay_Chat
;) Meybe another name would be more adequate.

producers/mimosa26_sbg/include/M26Producer.hh
```diff
@@ -0,0 +1,90 @@
...
+
+  void reset_com();
+
+  unsigned m_ev, m_run;
+  std::string m_verbosity, m_producerName;
+  bool m_terminated, m_configured, m_running, m_triggering, m_stopping;
+  eudaq::Configuration m_config;
+
+  // Add one mutex to be able to protect calls
+  //std::shared_ptr<unsigned> dummy;
```
simonspa: remove dead code

producers/mimosa26_sbg/include/irc_exceptions.hh
```diff
@@ -0,0 +1,173 @@
...
+ /** This exception class is the base class for all irc data exceptions
+   */
+  class DataException : public ircException {
+  public:
+    DataException(const std::string& what_arg) : ircException(what_arg) {}
+  };
+
+  /** This exception class is used in case a new event is requested but nothing available. Usually
+   *  this is not critical and should be caught by the caller. E.g. when runninng a DAQ with
+   *  external triggering and constant event polling from the DTB it can not be ensured that data
```
simonspa: Not sure you use a "DTB" here... ;)
CMS c&p

producers/mimosa26_sbg/include/irc_exceptions.hh
```diff
@@ -0,0 +1,173 @@
...
+  };
+
+  /**  This exception class is used when raw Mimosa values could not be decoded
+   */
+  class DataDecodingError : public DataException {
+  public:
+    DataDecodingError(const std::string& what_arg) : DataException(what_arg) {}
+  };
+
+  // not used
```
simonspa: remove dead code

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,511 @@
...
+
+using eudaq::RawDataEvent;
+
+
+M26Producer::M26Producer(const std::string &name,
+    const std::string &runcontrol,
+    const std::string &verbosity)
+  : eudaq::Producer(name, runcontrol), m_producerName(name), m_terminated(false), m_configured(false), 
+  m_running(false), m_triggering(false), m_stopping(false) {
+    m_sumEnDetectors = 0;
```
simonspa: why this variable isn't in the initializer list but separately in the actual execution block?

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+  } while (!m_terminated);
+}
+
+void M26Producer::OnConfigure(const eudaq::Configuration &param) {
+
+  m_config = param;
+
+  std::cout << "Configuring ...(" << m_config.Name() << ")" << std::endl;
```
simonspa: you do not guard your OnConfig with the mutex, so its call in MainLoop is useless.

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,511 @@
...
+	throw;
+      } catch (...) { 
+	printf("while connecting: Caught unknown exeption:\n");
+	//SetStatus(eudaq::Status::LVL_ERROR, "Stop error");
+	throw;
+      }
+
+      //m26_control->Init(m_config);
+      std::cout << " Initialising. " << std::endl;
+      try {
```
simonspa: probably just a style issue, but why do you wrap all of these calls in separate try-catch clauses?
if you fetch the mutex in one of them and the mainloop requests it, the loop will first be served before the next try-catch clause here will be executed. this might lead to problems since you just initialized the communication and then jump somewhere else.

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,511 @@
...
+	printf("while initialising: Caught exeption: %s\n", e.what());
+	throw;
+      } catch (...) { 
+	printf("while initialising: Caught unknown exeption:\n");
+	throw;
+      }
+
+      //m26_control->LoadFW(m_config);
+      std::cout << " LoadWF. " << std::endl;
+      try{
```
simonspa: see above

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,511 @@
...
+	printf("while loading FW: Caught exeption: %s\n", e.what());
+	throw;
+      } catch (...) { 
+	printf("while loading FW: Caught unknown exeption:\n");
+	throw;
+      }
+
+      //m26_control->JTAG_Reset();
+      std::cout << " Send JTAG RESET to sensors " << std::endl;
+      try{
```
simonspa: see above

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,511 @@
...
+
+    } catch (const std::exception &e) { 
+      printf("while Conf run: Caught exeption: %s\n", e.what());
+      throw;
+    } catch (...) { 
+      printf("while Conf run: Caught unknown exeption:\n");
+      throw;
+    }
+
+    /*try {
```
simonspa: remove dead code

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,511 @@
...
+}
+
+void M26Producer::OnTerminate() {
+  std::cout << "Terminate (press enter)" << std::endl;
+  m_terminated = true;
+
+  std::cout << " Unload FW " << std::endl;
+
+  try{
+
```
simonspa: mutex missing

producers/tlu/src/TLUProducer.cxx
```diff
@@ -42,7 +44,8 @@ class TLUProducer : public eudaq::Producer {
   void MainLoop() {
     do {
       if (!m_tlu) {
-        eudaq::mSleep(50);
+        //eudaq::mSleep(50);
```
simonspa: remove, not comment out. git history will contain it anyway.

producers/tlu/src/TLUProducer.cxx
```diff
@@ -302,6 +305,8 @@ class TLUProducer : public eudaq::Producer {
 };
 
 int main(int /*argc*/, const char **argv) {
+  std::cout << "Start TLU Producer \n" << std::endl;
```
simonspa: "Starting"?

cmake/FindThreads.cmake
arummler: I tried to compile the PR without FindThreads.cmake and it works fine.

producers/mimosa26_sbg/src/M26Producer.cxx
```diff
@@ -0,0 +1,431 @@
...
+    RunConf.TotEvNb          = 0; // only used when saving to disk
+    RunConf.EvNbPerFile      = 0; // noy used by DAQ
+    RunConf.FrameNbPerAcq    = 1800; // 1800 is optimum for M26, 400 for M28
+    RunConf.DataTransferMode = 3; // 0 = No, 2 = EUDET2, 3 = EUDET3 // will steer which data is send to shrd mem. 2 ->all frames, 3-> only w/ trig
+    RunConf.TrigMode         = 0; // deprecated
+    RunConf.SaveToDisk       = 0; // 0 = No, 1 = Multithreading, 2 = Normal
+    RunConf.SendOnEth        = 0; // deprecated, not need with eudaq
+    RunConf.SendOnEthPCent   = 0; // deprecated, not need with eudaq
+
+    sprintf ( RunConf.DestDir, "c:\\Data\\test\\1" );
```
arummler: Hard coded path

cmake/FindThreads.cmake
simonspa: Also on Windows? 'cause that will be our main target platform for this producer.

producers/mimosa26_sbg/src/M26Producer.cxx
simonspa: Data is not yet returned from the actual IPHC DAQ module, so for now basically everything is more or less a hack.
Before we finally merge this when things are working this obviously needs some cleanup  - but I wouldn't worry about it in this stage.

cmake/FindThreads.cmake
arummler: You are right, tested it only on my local linux machine. I will create my own PR and try it out on appveyor without that file.

arummler: Removing the file is enough on Windows as well.

arummler:
Please reopen against v1.7-dev

PR #245, opened 2016-06-09
EUDAQ interfacing LCIO via EUTel

tbisanz:
Updating various converter plugins to use the updated EUTelescope/LCIO interface

The (deprecated, see second PR) addSparsePixel() method now takes a const & instead of a *. Updated in all converter plugins which use this!

### DIFF
--- a/main/lib/plugins/APIX-CT-ConverterPlugin.cc
+++ b/main/lib/plugins/APIX-CT-ConverterPlugin.cc
@@ -240,8 +240,6 @@ namespace eudaq {
 	setupDescription.push_back( new eutelescope::EUTelSetupDescription( currentDetector )) ;
 	}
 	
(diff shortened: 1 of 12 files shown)


arummler:
Hi, if the interface changes, shouldn't we also check the eutelescope version?

simonspa:
that's broken anyway, I wouldn't bother there. And since it's @tbisanz
I'm pretty sure EUTel is taken good care of. ;)

arummler:
Just to make sure that a eudaq version with the adapted interface is not used with an old eutelescope version. But you are right, eutelescope should check for that if it uses eudaq as an external dependency.

simonspa:
Hej @tbisanz could you please re-open this against `v1.7-dev`? Thanks!

PR #261, opened 2016-07-06
pALPIDEfs: rotary stage integration

jvanhoor:
localized changes in palpidefs producer to include zaber rotary stages


simonspa:
Thanks, merged.

PR #262, opened 2016-07-08
Added EURunsplitter to the tools section

schuetzepaul:
Runsplitter divides large .raw files into several smaller files for parallel processing.
CMake flag: -DBUILD_runsplitter
Default: ON

### DIFF
--- a/etc/tools/runsplitter/CMakeLists.txt
+++ b/etc/tools/runsplitter/CMakeLists.txt
@@ -1,0 +1,12 @@
+SET(name "EURunsplitter.exe")
+
+SET(sourcefiles src/Runsplitter.cc)
+
+ADD_EXECUTABLE(${name} ${sourcefiles})
+
+target_link_libraries(${name} EUDAQ ${EUDAQ_THREADS_LIB})
+
+INSTALL(TARGETS ${name}
+  RUNTIME DESTINATION bin
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib)
--- a/etc/tools/runsplitter/src/Runsplitter.cc
+++ b/etc/tools/runsplitter/src/Runsplitter.cc
@@ -1,0 +1,151 @@
+/*
+ * Author: Paul Schuetze
+ *
+ * Purpose: Splitting .raw files from eudaq into several files
+ *
+ *
+ *
+ */
+
+
+#include "eudaq/FileReader.hh"
+#include "eudaq/FileWriter.hh"
+#include "eudaq/OptionParser.hh"
+
+#include <iostream>
+#include <string>
+#include <fstream>
+
+using namespace std;
+using namespace eudaq;
+
+void usage(){
+
+  cout << "EURunsplitter run options: " << endl << endl;
+  cout << "* -d (--directory) string\t\tDirectory containing raw data files" << endl;
+  cout << "* -o (--oldrun) int\t\tRunnumber of run to be splitted" << endl;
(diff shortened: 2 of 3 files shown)


arummler:
Nice tool, actually it is something I already needed myself. At some point I also wanted to add e.g. something to edit/create BORE/EORE.

etc/tools/runsplitter/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+SET(name "EURunsplitter.exe")
+
```
arummler: We did not change it yet for the 1.X branch but the explicite file ending should be removed.

etc/tools/runsplitter/src/Runsplitter.cc
```diff
@@ -0,0 +1,151 @@
+/*
+ * Author: Paul Schuetze
+ *
+ * Purpose: Splitting .raw files from eudaq into several files
+ *
+ *
+ *
+ */
+
```
arummler: In most files we are not doing this kind of comment header, style guide question? Reminds me: The Authors file should be updated anyhow.

etc/tools/runsplitter/src/Runsplitter.cc
```diff
@@ -0,0 +1,151 @@
...
+#include "eudaq/FileWriter.hh"
+#include "eudaq/OptionParser.hh"
+
+#include <iostream>
+#include <string>
+#include <fstream>
+
+using namespace std;
+using namespace eudaq;
+
```
arummler: It is not in a header, so it is still acceptable, but I prefer not having so (general) using...

etc/tools/runsplitter/CMakeLists.txt
schuetzepaul: Ok, didn't know that, sorry

etc/tools/runsplitter/src/Runsplitter.cc
schuetzepaul: Okay, i don't insist (or actually care) ;-)

etc/tools/runsplitter/src/Runsplitter.cc
```diff
@@ -0,0 +1,151 @@
...
+    sprintf(buffer, "%srun$6R$X", folder.c_str());
+  }else{
+    sprintf(buffer, "%s/run$6R$X", folder.c_str());
+  }
+  
+  string filename(buffer);
+  
+  FileReader * reader;
+  reader = new FileReader(origRunnr, filename);
+
```
arummler: line 78 and 79 can be combined, also try to avoid "naked" pointers

etc/tools/runsplitter/src/Runsplitter.cc
```diff
@@ -0,0 +1,151 @@
...
+      currentRunnr++;
+      
+      cout << "\nStarting new file with Run Nr.: " << currentRunnr << endl << endl;
+      
+      writer->StartRun(currentRunnr);
+      
+      log << "\t" << currentRunnr;
+      
+      // Write old BOREevent first
+      writer->WriteEvent(boreevent);
```
arummler: If the current time is available it might be nice to edit the BORE and also to insert something like these tags:
EditedAfterRun: True
EditedByTool(_X): EURunsplitter
ToolVersion(_X):1.0
ToolOutput(_X): Split at event ..., file x of y, Errors/Warnings: ...
EditDateTIme(_X):
Just an idea, there might be more elegant ways.

arummler:
I am doing some work on this - do not merge yet.

PR #308, opened 2016-11-23
V1.7 dev refactor c11

arummler:


tbisanz:
Just be careful with range-based loops, in most cases you will want a `(const) auto& ` and not merely `auto`. There is also the philosophical discussion if you should go for `auto&&` in these cases.

PR #310, opened 2016-12-05
Fixed include folders for new folder structure in EUtelescope

tbisanz:


simonspa:
Works as advertised, implementation is fine.

PR #313, opened 2016-12-09
Changing condition in cmake file for dl library

arummler:


main/lib/core/CMakeLists.txt
```diff
@@ -9,11 +9,12 @@ add_library(${PROJECT_NAME} SHARED ${CORE_SRC})
...
-  list(APPEND ADDITIONAL_LIBRARIES dl stdc++fs)
-  if(CMAKE_COMPILER_IS_GNUCC AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.3))
-    list(REMOVE_ITEM ADDITIONAL_LIBRARIES stdc++fs)
-  endif()
+if(CMAKE_COMPILER_IS_GNUCXX AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.3))
+  list(APPEND ADDITIONAL_LIBRARIES stdc++fs)
+endif()
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+  list(APPEND ADDITIONAL_LIBRARIES dl)
```
simonspa: It's better to use
```
LIST(APPEND ADDITIONAL_LIBRARIES ${CMAKE_DL_LIBS})
```
as this should be platform agnostic.

Present already in CMake 2.6, so no issues with that.

PR #317, opened 2016-12-16
Changes on the ITS Upgrade Code

freidt:
- Bug fix for the data decoding of pALPIDE-3 and ALPIDE
- Updates to the converter concerning the continuous-integration data analysis
- Producer only goes to error state if it cannot recover any more
- Updated the power supply control scripts


arummler:
Compilation fails under Linux (forgot to add changed header?) and Windows (root - win header conflict).

freidt:
@arummler : How can I update the version of the alice-its-alpide-software used by the CI? The changed header is actually part of the this external dependency.

arummler:
Currently the producer is compiled only under Linux and OSX as Windows would require some changes in alice-its-alpide-software (minimum makefile, die not look into details). For Linux and OSX it gets the zip file from the CERNBOX (see .travis/install_palpidefs_driver.sh) --> replace it in the CERNBOX and correct the path in the install...sh file. It might be necessary to empty the travis cache, I can do that in case you don't have the necessary access rights.

freidt:
1/8 builds failed due to exceeded job time limit. Amend and force push to trigger yet another rebuild.

arummler:
Hi, looks good and thanks for making it work. Just the VERSION variable could be named less generic. Merging now.

PR #318, opened 2016-12-16
Master fix ni

arummler:
Error in NIproducer: template and C++14 required (probably).


dreyling:
> Error in NIproducer
I think, you mean the NIConverterPlugin

arummler:
Yes, I corrected it now.

PR #349, opened 2017-03-22
Repaired nreader

gregorv:
nreader got broken a few months ago, these changes fix the compile- and linker-problems.


eyiliu:
Thanks for this fix

PR #357, opened 2017-05-24
Updated CMakeLists.txt to install headers

rete:
Hello eudaq devs !

I have updated to CMakeLists.txt files in main/lib/core and main/lib/lcio to install the header files when make install is triggered. Are you interested in ? This allows people to re-use the eudaq framework out of the box by using CMake to find the eudaq libraries and headers. I will produce a FindEUDAQ.cmake soon if you are interested in these changes and create a new pull request.

Cheers,
rete


rete:
OK, Travis seems to fail because there is a missing 'mkdir build' instruction in your .travis.yml file on the master branch.

arummler:
Hi,
I am working on improving the CI for the version 1 branch. As soon as I am done (time issues...) I wanted to adapt those for version 2 (master). Then it is going to work again (there are several things that went stale). Regarding the inclusion in other projects, e.g. eutelescope or pixelrce producer: This is a good idea but as far as I know FindSomething macros are meant for programs that do not use cmake as a build system. The documentation of cmake is a bit lacking and I am not a cmake wiz but to my undestanding a cmake based program should generate two eudaqConfig.cmake files for the installed and in tree build, plus a eudaqConfigVersion.cmake. Those are than installed and registered in the system, see e.g. the explanation in
https://cmake.org/cmake/help/git-master/manual/cmake-packages.7.html#creating-packages
I have a version 1 branch where I did already some work on installers for all our supported systems and trying to implement the generation of the mentioned files:
v1.7-dev-restructure-cmake-for-deployment

rete:
Hi !
I already produce this kind of files for DQM4HEP packages. Usually I do a copy-paste from a template and I adapt it for the target package. I can do this for EUDAQ if needed. The thing is that I'm working on a binding between EUDAQ and DQM4HEP. For this I'm developing a separate package (https://github.com/DQM4HEP/dqm4hep-eudaq) for which I need to load both packages from CMake script. That's why I need the EUDAQ headers. I can do something that works on my side and test it on Linux platform but I think it won't need some much modifications to work on other platforms. I'll produce something and push it asap.

PR #366, opened 2017-06-23
Workaround for xfer_bufptrs public/private bug in sstream header

arummler:
http://mailman.isi.edu/pipermail/ns-developers/2016-February/013378.html
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69470


arummler:
Combination clang 3.9 and root 5.34.36 does not work.

PR #367, opened 2017-06-24
CI: Adding cactus to linux and mac

arummler:


j1990grossmann:
I tried to compile version 1.6.0, which works out of the box, but with cactus not installed. Then in version 2.0.0. cactus framework has to be there. If someone needs eudaq mainly for a a telescope analysis there should be no need to have cactus installed. I don't know which of the BUILD flags has to be toggled to avoid installing cactus.

arummler:
You mean 1.7 right? As 1.6 is awfully outdated.
One flag is USER_CALICE_BUILD which enables USER_CALICE_BUILD_BIF. The corresponding converter plugin is built even if those producers are disabled. Another flag is USER_EUDET_BUILD_FMCTLU for the MiniTLU producer. The intention of this (unfinished) PR is to test those producers in the CI as well.

dreyling:
@arummler can we merge that?

arummler:
@dreyling In the moment the CI for both development branches as well as all OS is a bit of a mess. I have some private development and at some point I wanted to redo the whole thing with some different design choices. I would not merge it in the moment as this particular point I have to revisit.

PR #359, opened 2017-06-07
Develop

PaoloGB:
I have added to the commit the fil TLUhardware.cc that was missing in my previous merge request.
Apologies for the delay and for the mess.
Something happened to my repository and it took some time to sort out what was wrong.

PR #377, opened 2017-08-25
TLU v1c producer

PaoloGB:
Included modifications in the producer to cope with the new TLU data (6 words instead of 4).
Minor bug fixes.


arummler:
Dynamic load error handling in ModuleManager.cc only valid for Linux/Mac. Does not compile under Windows. Proposed solution:
#if EUDAQ_PLATFORM_IS(WIN32)
      //copied from https://mail.python.org/pipermail//pypy-commit/2012-October/066804.html
      static char buf[32];
      DWORD dw = GetLastError(); 
      if (dw == 0)
	      return NULL;
      sprintf(buf, "error 0x%x", (unsigned int)dw);
      //TODO get human readable https://msdn.microsoft.com/en-us/library/windows/desktop/ms679351(v=vs.85).aspx
      errstr = &buf;
#else
      errstr = dlerror();
#endif

arummler:
And does that mean that PR #359 can be closed?

arummler:
Is this necessary for the TORCH tb this week? In that case I will merge and try to patch.

Issue #383, opened 2017-08-31
euCliReader does not terminate cleanly ( segfaults )


DavidCussans:
Ran euCliReader over data written by Ex0TgDataCollector.

Segfaults at end.

Also segfaults if no events read ( specified with -e , -E )

```
[osboxes@osboxes bin]$ ./euCliReader -i 170831132543_run000017.raw -s -std -e 0 -E 2
<Event>
  <Type>2149999981</Type>
  <Extendword>171577627</Extendword>
  <Description>Ex0Tg</Description>
  <Flag>0x00000018</Flag>
  <RunN>17</RunN>
  <StreamN>324351285</StreamN>
  <EventN>0</EventN>
  <TriggerN>13836267</TriggerN>
  <Timestamp>0x0000000000000000  ->  0x0000000000000000</Timestamp>
  <Timestamp>0  ->  0</Timestamp>
  <Block_Size>0</Block_Size>
  <SubEvents>
    <Size>1</Size>
    <Event>
      <Type>2149999981</Type>
      <Extendword>3634980144</Extendword>
      <Description>TluRawDataEvent</Description>
      <Flag>0x00000011</Flag>
      <RunN>17</RunN>
      <StreamN>4008428646</StreamN>
      <EventN>0</EventN>
      <TriggerN>13836267</TriggerN>
      <Timestamp>0x0000035c64209abd  ->  0x0000035c64209ad6</Timestamp>
      <Timestamp>3695351732925  ->  3695351732950</Timestamp>
      <Tags>
        <Tag>BoardID=3209125689</Tag>
        <Tag>EUDAQ_CONFIG=Name = /home/osboxes/AIDA/eudaq/user/eudet/misc/fmctlu_runcontrol.conf

[Producer.fmctlu]
#AndMask = 0xf
#DUTInput3 = LEMO
#OrMask = 0
#PulseDelay = 31
#PulseStretch = 20
#TriggerInterval = 10000000
CLOCK_CFG_FILE = ./../user/eudet/misc/fmctlu_clock_config.txt
DACThreshold0 = -0.3
DACThreshold1 = -0.3
DACThreshold2 = -0.3
DACThreshold3 = -0.3
DutMask = 1
EUDAQ_DC = Ex0TgDataCollector
HDMI1_clk = 1
HDMI1_on = 0
HDMI2_clk = 1
HDMI2_on = 0
HDMI3_clk = 1
HDMI3_on = 0
HDMI4_clk = 1
[...]

PR #387, opened 2017-09-04
EUDAQ hard code changed: protect for EvtConverter if any empty evt po…

MengqingWu:
protect for empty event in NativeFileReader


eyiliu:
It is also related to the issue #383

PR #392, opened 2017-09-25
Sort hdmi

PaoloGB:
Sorted out the HDMI I/O lines selector.
Now it is possible to define the direction of each line (input/output) separately rather than on a connector basis.


PaoloGB:
Sorry, wrong pull request.

PR #395, opened 2017-09-29
TLU V1E Producer 

PaoloGB:
This is the code to run the TLU V1E producer.
There should be non conflicts. It builds and run on my machine but I still get an error when trying to run the euLog.


arummler:
Do you still need PR #359 ?

PaoloGB:
Not for now. That was for TLU V1C but it is unlikely that any user will need them.

arummler:
Okay, then it should stay open and be merged when it passes the checks.

PR #462, opened 2018-06-21
Merge in the user/aidastrip and user/tbscDESY

MengqingWu:
- local compile passed;
- update to the latest EUDAQ master branch;
- only user/aidastrip and user/tbscDESY added;
- 1 more change to the user/CMakeLists to add these two implementations;


MengqingWu:
@eyiliu Not sure why errors popped up. I have tested locally with 2 different machines, but of course the 2 added user both requires their specific environment to compile: 
1) user/aidastrip: it needs another daq headfiles and shared lib;
2) user/tbscDESY: it only needs mysql developement packages to be installed

And these users were switched to OFF by default for compilation.

eyiliu:
It is fine. The CI failure came from a known CI setup issue.

There are 2 data files, do you include them on purpose?
 user/aidastrip/exe/data/tracker_to_kpix_left.txt
 user/aidastrip/exe/data/tracker_to_kpix_left.txt

MengqingWu:
@eyiliu Yes the two files are supposed to be there to generate a header file for sensor mapping. I may need to update later, but it is fine for current setting.
Thanks a lot to let me know about the CI things :)

PR #480, opened 2018-08-10
Aida tlu naming convention: make file names and location more consistent

PaoloGB:
Following Jan's suggestion, I started making the naming convention for the new tlu more consistent.
All configuration and auxiliary files related to the AIDA TLU now have the previx _aida_tlu_.
There are located in
_/user/eudet/misc/hw_conf/aida_tlu_
I have not yet touched the actual source code files, but this is intentional: the configuration files are the ones most likely to be edited by users, so I think it is a good idea to implement this change as soon as possible. Likely, this will cause some disruption/confusion.

The producer is now called with

~./euCliProducer -n FmctluProducer -t fmctlu~
./euCliProducer -n FmctluProducer -t *aida_tlu*

The source code files should not affect users (except for the name of the producer).

Please let me know if you are not happy with this and we can define a better strategy.


PaoloGB:
I have almost finished renaming all the files belonging to the aida TLU.
I will create a merge for those instead.

PR #483, opened 2018-08-20
Improved logs and fixed bug.

PaoloGB:
Fixed:
- bug in address map for PulseT0 register (now 0x6 from 0x8).

Changes:
- the PulseT0 register has been replaced by RunActive; now it is necessary to specify the state of the TLU as running/stopped instead of simply pulsing the register. This was done to facilitate the shutter firmware change.
- tidied up log messages to make them more consistent. Now the verbose variable in the INI and CONF file affects the level of output log. A level of 1 is recommended.
- several logs have been "shifted" to the EUDAQ log so that they are not simply shown in the TLU producer.
- included "safety" checks: when some configurations are written, the code will check the read-back value to ensure it has been recorder correctly. If not, an EUDAQ error will be thrown. This should facilitate spotting misconfigurations, especially if the clock chip is not configured.


PaoloGB:
Added safety checks for illegal values in the CONFIG file.

PR #518, opened 2019-01-25
added some colors to logger cout and put the data/time first

lhuth:
just a minor style change just because Yi and myself had a chat about it


eyiliu:
The color message is nice on linux. However, windows can not handle the color message in the same way.  It might need to be reverted or improved on windows.

lhuth:
I'll have a look...

PR #532, opened 2019-03-07
AIDA: Add timestamps to StdPlane pixels & StdEvent

simonspa:
This PR extends the `StandardPlane` and `StandardEvent` classes by the possibility of storing timestamps. This is required for AIDA devices when using the EUDAQ-internal Raw->StdEvent conversions on the offline analysis.

IMHO the `StandardPlane` class should be completely rewritten, but that's a different story.

### DIFF
--- a/main/lib/core/include/eudaq/StandardEvent.hh
+++ b/main/lib/core/include/eudaq/StandardEvent.hh
@@ -30,35 +30,35 @@ namespace eudaq {
     virtual void Print(std::ostream & os,size_t offset = 0) const;
 
     /**
-     * @brief Funtion to retrieve begin of this event in nanosecons
+     * @brief Funtion to retrieve begin of this event in picosecons
      * @return Start time of this event
      */
-    double GetTimeBegin() const;
+    uint64_t GetTimeBegin() const;
 
     /**
-     * @brief Funtion to retrieve end of this event in nanosecons
+     * @brief Funtion to retrieve end of this event in picosecons
      * @return End time of this event
      */
-    double GetTimeEnd() const;
+    uint64_t GetTimeEnd() const;
 
     /**
-     * @brief Set begin time of event in nanoseconds
-     * @param begin Begin of events in nanoseconds
+     * @brief Set begin time of event in picoseconds
+     * @param begin Begin of events in picoseconds
      */
-    void SetTimeBegin(double begin) { time_begin = begin; };
+    void SetTimeBegin(uint64_t begin) { time_begin = begin; };
 
     /**
-     * @brief Set end time of event in nanoseconds
-     * @param end End of events in nanoseconds
+     * @brief Set end time of event in picoseconds
+     * @param end End of events in picoseconds
      */
-     void SetTimeEnd(double end) { time_end = end; };
+     void SetTimeEnd(uint64_t end) { time_end = end; };
 
     static StdEventSP MakeShared();
     static const uint32_t m_id_factory = cstr2hash("StandardEvent");
 
   private:
     std::vector<StandardPlane> m_planes;
-    double time_begin, time_end;
+    uint64_t time_begin, time_end;
   };
 
   inline std::ostream &operator<<(std::ostream &os, const StandardPlane &pl) {
--- a/main/lib/core/include/eudaq/StandardEvent.hh
+++ b/main/lib/core/include/eudaq/StandardEvent.hh
@@ -10,12 +10,12 @@
 
 namespace eudaq {
   class StandardEvent;
-  
+
   using StdEventUP = Factory<StandardEvent>::UP;
   using StdEventSP = Factory<StandardEvent>::SP;
   using StdEventSPC = Factory<StandardEvent>::SPC;
   using StandardEventSP = StdEventSP;
-  using StandardEventSPC = StdEventSPC;  
+  using StandardEventSPC = StdEventSPC;
 
   class DLLEXPORT StandardEvent : public Event {
     public:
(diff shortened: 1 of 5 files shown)


eyiliu:
Thanks to @simonspa  .
Is there any reason to use Double instead of Integral?

simonspa:
Hi @eyiliu,

the idea is the following:

* `Event` contains timestamps in (whatever) bins of the hardware
* `StdEvent` contains `double` time in units of nanoseconds

The basic problem here is, that if we want to use the `Raw2StdEvent` converter as a generic converter, the data returned in `StdEvent` has to be in universal units. If it still is in some (unknown) clock ticks, I have to translate that in user-code on a per-detector basis into a common time binning.

This is why I decided to implement it as `double` in nanoseconds, so the data returned by `StdEvent` is common across all detectors.

eyiliu:
Hi @simonspa , 

* Unfortunately, there is no generic Raw2StdEvent convert. The bin timestamp in RawEvent is filled by  per-Producer which must anyhow partly decode the hardware data to extract trigger and timestamp information.   The only chance to have common time binning is in the Producer on per-detector basis.

* the **uint64** by now is long emough to keep a common timestamp in nanosecond precision (start from  **1970 Epoch time**) .

*  std::chrono::nanoseconds is also 64bit integral internally .

simonspa:
Hi @eyiliu 

I am well aware that there is no generic converter - that's not what I meant. What I mean is a software, that uses the EUDAQ converter factory to do e.g. this:

```cpp
auto stdevt = eudaq::StandardEvent::MakeShared();
eudaq::StdEventConverter::Convert(evt, stdevt, nullptr))
auto my_time = stdevt->GetTimeBegin();
```

We need to make sure that the returned time is always given in the same units. Settling for `uint64_t` or another integer type is not a good idea in my opinion because this limits the precision of the clock. If I have a device producing pico-second timestamps, I'm lost already. If it produces 12.5ns-binned timestamps, I'm also lost because I get rounding issues of half a nanosecond. Etc etc. It's not about worrying about the epoch of the timestamp but about its **precision**.

This is why I think using a double here is both justified and sensible, and we should just settle (and document) the base unit of the values.

eyiliu:
Hi @simonspa 

How do you define the Timestamp Zero? As my understanding, a double could provides 53bits integral internally. If there is a picosecond-wise detector using double, 53bit can only cover 9007199254740991 which is 150 minutes in picosecond granularity if Timestamp Zero is set as the start of data-taking. The situation is even worse if the Zero is setup to a point earlier. I think it is reason why generic implement, eg. time_t and std::chrono::time_point do not use double.

I would suggest to have integral base on picosecond unit for the picosecond-wise detector. It brings the fully using of  64bits instead of 53bits


Yi

simonspa:
But you cannot have different base units for different detectors - that completely defies the idea of having a common decoding interface. In my above code snippet, I need to know what the unit of the value returned by `GetTimeBegin()` is, independent of the detector type behind it (because at this time, O don't even know about different detectors!).

And with

```bash
std::numeric_limits<double>::max()    =  1.79769e+308
std::numeric_limits<double>::lowest() = -1.79769e+308
std::numeric_limits<double>::min()    =  2.22507e-308
```

we should be covered. Or what am I missing?

eyiliu:
Hi @simonspa 

The advantage of the double is it can dynamic covering a huge range.
> std::numeric_limits<double>::max() 
> std::numeric_limits<double>::min() 

However, when huge range is covered, the precision is worse.  The 53bits provide about 15 decimal digital precision.

It means that
the double will not say the difference of
1.23456789012345**77**E16
1.23456789012345**11**E16


Yi

simonspa:
True but probably a corner case for us with 1e16ns ~ 100 days.

lhuth:
Hi @simonspa, 
I think, @eyiliu  is concerned about the fact, that double provides a scale dependent absolute precision - for numbers in the order of 1e16, the precision will be in the order of 1ns. In contrast, the precision of a int is always +/-1 bit.  In a normal testbeam campaign this is unlikely to cause any trouble.
However, @eyiliu seems to prefer integer values in order to be immune against this effect as he is concerned that people using for example pico-seconds timestamp are not aware of it and are loosing precision after 2.4 hours of data taking starting at zero.

Do you really need  double time_ns() or would be let's say int time_ps() also sufficient? I think this might be a reasonable? compromise, as I am not aware of a system with timestamps below one pico-second.  However, this is not thinking in a long term view...

simonspa:
In the end I don't care - all I know is that I need a stable reference time retrievable from the EUDAQ2 converter factories, since all our reco builds on using timestamps to sync. Having a large if-else block with detector-dependent conversion factors in a thing called `EventLoaderEUDAQ2` does not seem too intriguing to me.

eyiliu:
Hi @simonspa 

* I agree with your point to have common timestamp unit. If-else block on per-Event-type case is neither nice nor friendly to user. 
* Double can not provide sufficient precision on 1ps when data-taking time goes longer than 2.4 hours no matter whether the ns or ps unit is used.
* The ps is definitely better than ns to provide fine precision if integral is considered to be used.  

If the above 3 points are correctly and you agree with,  the solution could be to have   
> uint64_t GetTimestamp_ps()

simonspa:
I'd be fine with that (apart from the name which could probably adhere a bit more to the rest of the EUDAQ code).

We need this for `StdEvent` and for `StdPlane` classes. And as noted in a commit message somewhere: the latter should be rewritten completely - it's a mess.

eyiliu:
Yes, the StdPlane is mess and need to be rewritten. As you said, it is another topic and task. 
The function name was just example, please give it a proper name.

When you change to ps unit, please help to change TLU unit, now ns, to ps in the same PR. 
Thank you in advance. 

https://github.com/eudaq/eudaq/blob/master/user/tlu/module/src/EudetTluProducer.cc#L115
> ev->SetTimestamp(ts_ns, ts_ns+8, false);

https://github.com/eudaq/eudaq/blob/master/user/tlu/module/src/AidaTluProducer.cc#L80
> ev->SetTimestamp(ts_ns, ts_ns+25, false);

simonspa:
Hi @eyiliu 

I switched this in favor for `uint64_t` in picoseconds. The two lines you quite in your last comment do not need to be changed since they only set the timestamp in the `Event` class - where it can be whatever the detector wants. Only in the converted `StandardEvent` it has to be in picoseconds.

I will take care of this for the TLU.

/Simon

simonspa:
Now done, should be ready for review. Thanks.

eyiliu:
Nice to have.  Reviewed. Ready for merging?

simonspa:
Ready from my side.

eyiliu:
Waiting for CI check finished

simonspa:
AppVeyor failed due to timeout in AFS installation, all other builds passed.

lhuth:
AppVeyor is frequently failing for at least one of the builds - nothing to worry about.

PR #536, opened 2019-03-18
Correctly propagate the result of event decoding

simonspa:
Right now, event converter cannot properly fail because the returned `false` boolean from the event converter is not propagated by the `RawEvent2StdEventConverter` class.

This PR fixes that.


simonspa:
Thanks to @jekroege for digging and fixing.

eyiliu:
Thanks @jekroege and @simonspa

PR #538, opened 2019-03-21
Revert scan features

lhuth:
Yesterday I found several commits from myself in the master branch that should not be there yet. 
I reverted these changes.
Unfortunately I do not know how these features ended up in the master.

As reverting parts is something I didn't need to do anytime before please check this request carefully before merging it.  It compiles and the example scripts are running on linux.

### DIFF
--- a/gui/include/euRun.hh
+++ b/gui/include/euRun.hh
@@ -57,8 +57,6 @@ private:
   bool checkScanParameters();
   void createConfigs();
   bool allConnectionsInState(eudaq::Status::State state);
-  bool checkEventsInStep();
-  int getEventsCurrent();
 
   static std::map<int, QString> m_map_state_str;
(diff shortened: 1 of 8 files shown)


dreyling:
I have checked it and did one reversion more, since there was no data-channel to the DataCollectors. Basically, now the following files are now at the status from 18.  Feb from commit  58330b79b0d63b728e38777865318b2205959379
```
gui/include/euRun.hh
gui/src/RunControlModel.cc
gui/src/euRun.cc
gui/ui/euRun.ui
main/lib/core/include/eudaq/Logger.hh
main/lib/core/include/eudaq/RunControl.hh
main/lib/core/src/Configuration.cc
main/lib/core/src/RunControl.cc
```
Again sorry for probably my mess-up @eyiliu and @lhuth -- now we can pick later developments of these files.

PR #545, opened 2019-04-04
added scan functionality and documentation of the GUI

lhuth:
Introduced a class to handle the scanning of parameters.
The scanning functionality required some major changes in the gui layout.
However, the eudaq/RunControl has nto been touched and all changes/features are realized on the gui level


lhuth:
I've added another state to the EUDAQ2 state machine called "stopped", which allows to stop - configure - start in a controlled way without resetting everything which is rather time consuming.

lhuth:
Moved to development branch PR (v2-dev)

PR #551, opened 2019-05-14
Scan PR with addtional RC State (Stopped)

lhuth:
-> Scanning features
-> Documentation
-> RC State stopped. Allows reconfigure without reset -> faster better stable :)
-> Serveral minor fixes


dreyling:
It was successfully tested with NiProducer and EudetTluProducer and the corresponding hardware and all other components on Linux.
A pre-version was already successfully applied at a test beam of the CLIC group.

PR #560, opened 2019-06-24
Added the scanning features

lhuth:
Tested in beam for several hours without issues

added the possibility of parameter scans.
This has some implications:
- `RunControl` has an additional state `STOPPED`, which allows safe reconfiguration without reset
- `GUI`: Added documentation and a helper class to control the scan


gui/include/euRun.hh
```diff
@@ -39,14 +43,26 @@ private slots:
...
+
 private:
+  eudaq::Status::State updateInfos();
   bool loadInitFile();
   bool loadConfigFile();
   bool addStatusDisplay(std::pair<eudaq::ConnectionSPC, eudaq::StatusSPC> connection);
   bool removeStatusDisplay(std::pair<eudaq::ConnectionSPC, eudaq::StatusSPC> connection);
   bool updateStatusDisplay();
-  bool addToGrid(QString objectName, QString displayedName="");
+  bool addToGrid(const QString objectName, QString displayedName="");
```
eyiliu: ```const QString &```  reference is more reasonable

gui/include/scanHelper.hh
```diff
@@ -0,0 +1,143 @@
...
+#define EUDAQ_INCLUDED_GUI_ScanHelper
+
+#include <stdio.h>
+#include <string>
+#include <vector>
+#include "eudaq/Config.hh"
+#include "eudaq/Logger.hh"
+#include "eudaq/RunControl.hh"
+
+using namespace std;
```
eyiliu: no default ```namespace std``` at header file.  It will affect all the other files which include this ScanHelper.hh

gui/include/scanHelper.hh
```diff
@@ -0,0 +1,143 @@
...
+     */
+    string currentCountingComponent() const;
+private:
+
+    /**
+     * @brief ERROR message from EUDAQ_ERROR with return statement
+     * @param msg
+     * @return false
+     */
+    bool ERROR(string msg) {    std::cout << msg <<std::endl; EUDAQ_ERROR(msg); return false;}
```
eyiliu: all up-case  ERROR is more like a pre-processing macro than function. Suggest to have some name else.

gui/include/scanHelper.hh
eyiliu: No std in header file

main/lib/core/src/RunControl.cc
```diff
@@ -57,32 +57,11 @@ namespace eudaq {
...
-	    + ":"
-	    + server_addr.substr(server_addr.find_last_not_of("0123456789")+1);
-	}
-	std::string server_name = conn_type+"."+conn_name;
-	if(server_name=="LogCollector.log" && !server_addr.empty()){
-	  m_conf_init->SetSection("");
-	  m_conf_init->SetString("EUDAQ_LOG_ADDR", server_addr);
-	  SendCommand("LOG", server_addr);
-	}
-      }
```
eyiliu: Do you have test run with Logger? I guess the remove part is needed.

gui/include/scanHelper.hh
lhuth: true - remnant from testing

gui/include/scanHelper.hh
lhuth: done

main/lib/core/src/RunControl.cc
lhuth: It's tested and running - I just used the  InitialiseSingleConnection() function to not duplicate the code

gui/include/scanHelper.hh
lhuth: same file as above - changed

gui/include/euRun.hh
lhuth: done

simonspa:
Any estimate on merging this? We would like to add that to our stack and get it tested before our next beam time.

user/example/misc/scan/ExampleScan.scan
```diff
@@ -0,0 +1,27 @@
+[global]
+repeatScans = 1
+allowNested = 0 # overrides local nested arguments
+configPrefix = "/home/lhuth/test/scanConf/scanned"
```
eyiliu: Is it possible be  a relateive path?

eyiliu:
Schedule to be merged today or tomorrow. @simonspa

user/example/misc/scan/ExampleScan.scan
lhuth: It's up to the user to set the path to be relative - this makes it, however, execution folder dependent and I do not recommend to do so.

PR #561, opened 2019-07-03
Add Caribou Producer & Renovate Timepix3 Producer

simonspa:
This MR adds a new producer for the Caribou DAQ system. The producer works with any detector implemented in the Caribou Peary DAQ software and runs directly on the System-on-Chip CPU of Caribou.

Currently, there are data converters for two devices (CLICPix2, ATLASpix) and new devices can simply register their data type and add a converter.

Furthermore, the Timepix3 producer using the SPIDR system has been completely revamped and now works in the AIDA mode, synchronizing with TLU clock and T0.


simonspa:
@lhuth @jekroege

simonspa:
I accidentally introduced the cast in #532 not realizing that it is unnecessary.

PR #562, opened 2019-07-04
StandardEvent: initialize event start/end with zero

simonspa:
We need to initialize event start and end timestamps to zero, otherwise they might contain random memory content.


eyiliu:
Thanks for the enhancement.

PR #565, opened 2019-07-11
Python 3 compatibility for Python EUDAQ Wrapper

DavidLP:
To be able to use BDAQ53 with EUDAQ the Python wrapper needs Python 3 compatibility. This is added with this PR. The functionality of this wrapper with Python 3 is successfully tested within the [BDAQ53 CI pipeline](https://gitlab.cern.ch/silab/bdaq53/-/jobs/4793744).

### DIFF
--- a/python/PyEUDAQWrapper.py
+++ b/python/PyEUDAQWrapper.py
@@ -1,12 +1,23 @@
 import sys
-from ctypes import cdll, create_string_buffer, byref, c_uint, c_void_p, c_char_p, c_size_t, c_uint8, POINTER,c_long
+import ctypes
+from ctypes import cdll, byref, c_uint, c_void_p, c_char_p, c_size_t, c_uint8, POINTER, c_long
 import numpy
 import os.path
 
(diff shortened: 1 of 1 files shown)


arummler:
Our CI failed to build due to CI reasons (I am actually working on it). But before I merge: is it still compatible with Python2? Or should we have a separate python3 wrapper?

DavidLP:
The intention was to keep it Python 2 compatible and the code changes are small. But honestly, I did not check for Python 2 compatibility. We are about to (as pretty much [all famous Python libraries](https://docs.scipy.org/doc/numpy-1.14.2/neps/dropping-python2.7-proposal.html)) drop Python 2 support.

arummler:
Fine with me, I will merge.  Part of my (unfortunately very slowly ongoing) CI refurbishment is that I am trying to build a wheel for all three systems as it was requested at some point and to upload to pypi but I am not there yet. Might need some check from actual python eudaq users at some point in the future.

Issue #572, opened 2019-07-31
Light-weight installation for reconstruction?


jekroege:
It would be very useful to have something like a light-weight installation of eudaq2 that only compiles the `StandardEvent converters`.

The problem I'm facing:
When I'm running my reconstruction using `corryvreckan`, I need a full installation of eudaq2 to be linked against which means that I also need to install `cactus` and all the other packages which are needed for the TLU for instance.
However, these packages should only be required when actually taking data with eudaq, not to compile the converters.

This also makes it particularly complicated to get the reconstruction (with corryvreckan) to run on a new PC as I have to install many many packages.

jekroege:
It would be very useful to have something like a light-weight installation of eudaq2 that only compiles the `StandardEvent converters`.

The problem I'm facing:
When I'm running my reconstruction using `corryvreckan`, I need a full installation of eudaq2 to be linked against which means that I also need to install `cactus` and all the other packages which are needed for the TLU for instance.
However, these packages should only be required when actually taking data with eudaq, not to compile the converters.

This also makes it particularly complicated to get the reconstruction (with corryvreckan) to run on a new PC as I have to install many many packages.

lhuth:
Hi @jekroege,
thanks for pointing that out. 

The desired behaviour is already implemented: If you set `USER_TLU_BUILD=ON` the converter will be build even if the external dependencies (`cactus`, `ZestSC1` and `libusb`)are not found. 

However the cmake output is misleading and we should change that as discussed online.

jekroege:
All done, see MR #573

eyiliu:
Thanks @jekroege for #573

PR #573, opened 2019-07-31
Improve CMAKE, allow compilation of converters only

jekroege:
This solves #572.

As discussed with @lhuth, the converter for the TLU is being built, even if CACTUS or ZESTSC1 are not found. The cmake output however, is misleading. This is fixed in this merge request.

In addition, also the Timepix3 Converter can now be built when xerces-c or SPIDR are not found. This is only required for config file parsing in `Timepix3Config.cxx`.
Also the ROOT requirement is removed as not needed.


lhuth:
I am happy with it - nice and quick :)

Issue #576, opened 2019-08-28
Way to determine detector from RaqDataEvent?


simonspa:
We need a way to determine the detector data contained in a UEDAQ2 RawDataEvent, preferrably before decoding. However, there does not seem to be a coherent way to do so.

As far as I can see, we have the following information available:

```cpp
auto my_raw_event = std::make_shared<eudaq::Event>();
// Returns name of the raw event type the factory is registered with, e.g. "NiRawData"
// The type we get here does not correspond to the detector (M26) and is confusing
//  to users - because normally they don't have to deal with these "EUDAQ2-internal"
// names
std::cout << my_raw_event->getDescription() << std::endl;

auto my_std_event = std::make_shared<eudaq::StandardEvent>();
eudaq::StdEventConverter::Convert(my_raw_event, my_std_event);
// Still the same cryptic name
std::cout << my_std_event->getDescription() << std::endl; 

// For events without planes, here we are stuck (e.g. TLU) there is no way to query
// anything in the event that would return "tlu".

// For other events, we can loop over all planes:
for(size_t i_plane = 0; i_plane < my_std_event->NumPlanes(); i_plane++) {                                                                        
    auto plane = my_std_event->GetPlane(i_plane);
    // Now we finally get something that looks like an identifier for the detector
    std::cout << plane.Sensor() << std::endl;
    // - but there might still be multiple planes, so let's check
    std::cout << plane.ID() << std::endl;
    // which is just an enumerator as it seems.
}
```

Could we somehow provide a better interface for other software? We need to be ab;le to determine which detectors are stored in a raw event to select which events we should be taking into account for the reconstruction.

simonspa:
@jekroege

lhuth:
Hi @simonspa,

Sorry for the rather late feedback - been on holidays.

The naming is somewhat unfortunate, I agree. However, changing this disables reading old telescope data. 

RawEvents do not know what data they contain - it's simply blocks and a very basic header.  

In your above example, you are converting an `Event` to a `StandardEvent`. The converison only "unpacks" the data blocks and adds this information. I think this could be rather simply improved by adding something like `StandardEvent->SetDetector(std::string name)` and  `StandardEvent->GetDetector()`. As the converters can interpret the blocks, they could set this name. However, this assumes, that only one type of detector is stored in an event, which is maybe not always the case. And it also requires event conversion first, which slows down reading.

We are open for better ideas.

simonspa:
[...]

PR #578, opened 2019-09-05
Allow retrieval of detector type from StdEvent

simonspa:
This is a PR intended to solve the issue raised in #576 

It adds a new member plus setter/getter to StdEvent to store the detector type. It has to be set explicitly by the StdEventConverter of the respective detector.


simonspa:
I would have loved to shield all the setters from the outside through friendship, but since friendship is not inherited, that doesn't work without adding boilerplate code to `eudaq::Dataconverter` or `eudaq::StdEventConverter`.

PR #583, opened 2019-09-20
Add proper CMake Targets

### LINKED ISSUE #582: EUDAQ2 CMake Targets not complete
Currently, EUDAQ2 only exports the `eudaq_core` target. However, to really make use of this in dependent packages, we would also need targets for the EUDAQ2 modules created in the build, along with their dependencies.

simonspa:
This PR adds CMake targets for all modules as well as an interface library which encompasses all modules. In dependent projects, EUDAQ2 can then simply be linked by using

```cmake
FIND_PACKAGE(eudaq REQUIRED)

ADD_LIBRARY(MyLib SHARED MyLib.cpp)
TARGET_LINK_LIBRARIES(MyLib PUBLIC eudaq::core eudaq::modules)
```

without the need to juggle with include and library dirs.

This fixes #582


simonspa:
Since this is a required change for other frameworks to interface EUDAQ I'd be grateful for having a new tagged version 2.4.0 which we can then set as requirement.

simonspa:
Ready for review.

eyiliu:
To be consistent with current CMake style in project, low case cmake-commands are preferred.  
Thanks

eyiliu:
The CMake interface feature is very nice! Thanks.

1. For a normal case,  user module is only loaded by the core at runtime and does not export anything. Therefore, a normal user module will be not linked (being the dependency) by any other.

2. The lcio external lib (libeudaq_lcio.so) might need to enjoy the new interface family.

simonspa:
1. In the reconstruction software I need to link against all user modules in order to enjoy the Converters registered to the factory - or am I wrong>
2. True, will add this

simonspa:
Done. The CMake syntax capitalization style seems pretty mixed from what I have seen. You want me to redo them all?

eyiliu:
@simonspa 
> The CMake syntax capitalization style seems pretty mixed from what I have seen. You want me to redo them all?
Of cource, I'm not asking to do all of them. If you modified a CMake file, somehow to be consistent
to other part of this file is reasonable. There are some mixup. The main part of the eudaq is lower case.

eyiliu:
>     1. In the reconstruction software I need to link against all user modules in order to enjoy the Converters registered to the factory - or am I wrong>

No, if you are using the converter offerred factory.

simonspa:
I changed all CMake commands of the main library and CMakeLists to lower-case. Ready for review/merge.

CMakeLists.txt
```diff
@@ -1,11 +1,18 @@
 project(EUDAQ)
 set(${PROJECT_NAME}_VERSION_MAJOR 2 )
-set(${PROJECT_NAME}_VERSION_MINOR 0 )
+set(${PROJECT_NAME}_VERSION_MINOR 4 )
```
lhuth: Version 2.3.0 currently, right?

user/aidastrip/module/CMakeLists.txt
```diff
@@ -24,10 +24,13 @@ target_link_libraries(${EUDAQ_MODULE} ${EUDAQ_CORE_LIBRARY}
 set_target_properties(${EUDAQ_MODULE} PROPERTIES INSTALL_RPATH
   ${EUDAQ_INSTALL_RPATH}:/usr/local/lib/kpix/)
 
+# Add to modules library:
+TARGET_LINK_LIBRARIES(modules INTERFACE ${EUDAQ_MODULE})
```
lhuth: same here

user/calice/module/CMakeLists.txt
```diff
@@ -15,8 +15,12 @@ add_library(${EUDAQ_MODULE} SHARED ${MODULE_SRC})
 target_link_libraries(${EUDAQ_MODULE}
   ${EUDAQ_CORE_LIBRARY} ${EUDAQ_LCIO_LIBRARY} ${LCIO_LIBRARIES})
 
+# Add to modules library:
+TARGET_LINK_LIBRARIES(modules INTERFACE ${EUDAQ_MODULE})
```
lhuth: same here

CMakeLists.txt
simonspa: Yes, I want to tag v2.4.0 immediately after merge because that will be a change how to interface EUDAQ

lhuth: Okay.

simonspa:
@eyiliu any chance to get this merged soon?

simonspa:
Thank you for merging, @eyiliu. Could we also have tag v2.4.0 so we can require that version when linking against it?

eyiliu:
done

Issue #543, opened 2019-03-28
Adding DUT plots to the EUDAQ Online Monitor ?


yotarid:
Dear EUDAQ experts, 

I would like to know if it is possible to add DUT plots to the EUDAQ Online Monitor ? (nothing related to that is in the EUDAQ manual)

If it is the case, is there any existing documentation that describes how to do it ? 

Thanks in advance for you answer,

Best regards
Younes

arummler:
First a little disclaimer: The code of the online monitor is a bit messy and will be hopefully re-written in the future. If you implement the standard event conversion in your converter plugin then your DUTs should appear automatically. It provides xy-positions, ToT, LVL1, correlation plots. If you need a plot that is very specific to your detector you have to add it to the online monitor first.

yotarid:
Sorry for the late answer. We actually want to add some very specific plots (amplitude spectrum, FWHM histogram, ...). Is there any documentation that shows how to do it ? Or maybe you can point me to the code parts that I should work on ?

Issue #337, opened 2017-03-09
FileWriterRoot


hoide:
I was trying to use FileWriteRoot in the converter (v1.5.0) and found that the output ROOT file is broken (says "has no keys"):
`main/lib/src/FileWriterRoot.cc`

Looking at the code, I think the following lines should be added.
* in `void FileWriterRoot::StartRun(unsigned runnumber)`:
```c++
m_ttree->AutoSave("SaveSelf");
```
* and also the following line in `FileWriterRoot::~FileWriterRoot()`
```c++
m_tfile->Write();
```

Best,
Hide

arummler:
Please try the current v1.7-dev development branch. There were some changes in the converter.

arummler:
Did you try the more recent version?

PR #585, opened 2019-10-04
ROOT monitor

forthommel:
This PR adds a highly customisable ROOT-based monitoring tool to the `eudaq` universe.

It has been developed in the scope of the 2019 CMS-TOTEM test beams campaigns at DESY-II. For this latter, an example of such an implementation including low-level detector monitoring plots may be found e.g. [in here](https://github.com/forthommel/eudaq/blob/cmstotem-dev/user/cmstotem/module/src/SampicMonitor.cc).

The build may be triggered adding `-DEUDAQ_BUILD_ONLINE_ROOT_MONITOR=ON` to the CMake configuration flags.

~Additionally, a "bug" fix is introduced in the `TTreeEventConverter` utilitary that prevented the proper filling of the output tree (or lead to corrupted tree content).~


arummler:
Interesting work, using plugins goes into a direction that I wanted to try with the current OnlineMon to address its major problems: hard coded decision paths for different DAQs which make it inflexible and cumbersome to maintain. Furthermore, the standard event which fits only the lowest common subset of strip/pixel detectors (already ToT is only implemented as a hack). The idea was to drop potentially standard event totally and make dedicated plugins for each application (monitor, converter, whatever). I did not have yet time to look thoroughly through your code but I see one issue that I had also thought about and where I did not really find an elegant solution yet: correlation plots. In tracker test beams one wants correlations between basically all devices and sometimes 3-4 different producers. In your SAMPIC plugin you are accessing the plane 0 of the telescope in a hard coded way which is a similar application but not really feasible for universal correlation plots. Probably one could extend it by a mechanism where a plugin can provide data to other plugins and announce its capabilities... Not sure yet.

Other thing that I tried to study is the speed. The current OnlineMon is quite slow, both for online as well as offline usage which forced us to introduce the skipping of events. I put the eudaq1 OnlineMon through a profiler and the filling of the root histograms dominated (because there are typically a lot). One "brute force" approach which I wanted to try was to turn the monitor itself into a server which could run distributed over several cores/machines and deliver the histograms to web clients using rootjs. This could be tried on top of your work, would probably go along with the highly distributed philosophy of eudaq2.

PS In your converter plugin I saw a bunch of "magic numbers", probably it would be better to make constants. I will try to take a more detailed look later.

arummler:
I am not sure about the name as the stdevent monitor is also using root for drawing (and the offfline monitor in eudaq1 uses it as well). Also it should be added to the documentation.

arummler:
The fail of the Mac CI is currently a CI isssue. I will look into it.

arummler:
Please merge #586 and re-trigger CI for this one.

forthommel:
Hi @arummler,

Many thanks for this very prompt and detailed review (including the foreseen PR to appear including our own developments, which still requires a major cleanup before its submission to this repository!)

I must say I was also prepared for a bit of discussion over the naming of this new intermediate object. Whatever proposal that would suit the eudaq conventions is of course very much appreciated.
For the correlations plots, it would indeed be feasible to implement a few helper classes to allow the variables definitions and correlations. I was thinking of something defined along [pseudo-code]:
```cpp
template<typename Obj1,typename Obj2=Obj1> TH2* // <- or whatever object... TProfile2D, ...
ROOTMonitorWindow::BookCorrelationVariable(
    const std::string& path, 
    const std::function<double(void)>& var1 = std::bind(&Obj1::var1, this),
    const std::function<double(void)>& var2 = std::bind(&Obj2::var2, this));
```
that could allow of any float-returning member method to be used independently (assuming a converter is provided for the template arguments objects). But of course any implementation would be possible there...

forthommel:
@arummler FYI, I started preparing a manual for this new object. See https://github.com/forthommel/eudaq/compare/root-monitor...forthommel:root-monitor-manual

forthommel:
With https://github.com/eudaq/eudaq/pull/585#commits-pushed-96598cb I also included the capability to open in the monitoring tool:
* either monitoring .root output files,
* or more importantly, raw data! (where vistars are created and populated equivalently as during the online monitoring).

lhuth:
HI @forthommel,
thanks a lot for the PR - Is it ready for review or still WIP? 
Our main EUDAQ2 developer (@eyiliu ) is currently on vacation and will be back on monday. Therefore, we will start reviewing your changes next week.
It would be great if you would inform us if the PR is finalized.
Cheers,
Lennart

forthommel:
Hi @lhuth !
After a bit of fighting with the Darwin part of the CI I believe it is now ready for review. The only remaining part is this correlations definition environment, which may be the topic of a follow-up PR...

Cheers, Laurent

forthommel:
A few utilitary plots useful for any user-defined monitor were added in https://github.com/eudaq/eudaq/pull/585/commits/5c04f762ab2910d736526383fb7b52602ef3fa0d.
Beside bug fixes no new features are expected prior to your reviewing of this PR.

monitors/rootmonitor/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+  message(STATUS "Online ROOT monitor is not to be built on windows")
+  return()
+endif()
+
+list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
+if(WIN32)
+  find_package(ROOT REQUIRED COMPONENTS libGui libCore libHist libRIO)
+else()
+  find_package(ROOT REQUIRED COMPONENTS Gui)
+endif()
```
lhuth: Maybe you could be a bit more consistent here: If you do not support windows at all, the if-else in 7-11 is not necessary. Otherwise, you can remove lines 1-4.

monitors/rootmonitor/include/eudaq/ROOTMonitorWindow.hh
```diff
@@ -0,0 +1,175 @@
...
+/// A placeholder for ROOT monitoring
+class ROOTMonitorWindow : public TGMainFrame {
+  static constexpr const char* NAME = "ROOTMonitorWindow";
+  RQ_OBJECT(NAME)
+public:
+  ROOTMonitorWindow(TApplication*, const std::string& name);
+  ~ROOTMonitorWindow();
+
+  /// List of FSM states
+  enum class Status { idle, configured, running, error };
```
lhuth: EUDAQ2 is already shipped with a set of states - is there a reason not to reuse them?

monitors/rootmonitor/src/ROOTMonitor.cc
```diff
@@ -0,0 +1,132 @@
...
+
+namespace eudaq {
+  ROOTMonitor::ROOTMonitor(const std::string & name, const std::string & title, const std::string & runcontrol)
+    :Monitor(name, runcontrol),
+     m_app(new TApplication(name.c_str(), nullptr, nullptr)),
+     m_monitor(new ROOTMonitorWindow(m_app.get(), title)){
+    if (!m_monitor)
+      EUDAQ_THROW("Error allocating main window");
+
+    m_monitor->SetStatus(ROOTMonitorWindow::Status::idle);
```
lhuth: A new instance of any EUDAQ component will start up automatically in an 'uninit' state. I do not think you'll need that state here.

monitors/rootmonitor/CMakeLists.txt
arummler: If it is supposed to become the base for a new standard monitor to replace OnlineMon (eudaq1) / StdEvtMon (eudaq2) then one should try to get it running on all our standard supported platforms. What problem did you encounter with VS on Windows?

forthommel: @lhuth, @arummler 
Thanks for the review! I indeed forgot I disabled this on win systems (pretty much like this is done in the `StdEvtMon`, see [here](https://github.com/eudaq/eudaq/blob/master/monitors/onlinemon/CMakeLists.txt#L1-L4)), as I did not have a working environment to test these developments on.
I will try to set up a VM with Win/VS for further tests (and re-enable the compilation for the time being to see if the CI complains...)

monitors/rootmonitor/include/eudaq/ROOTMonitorWindow.hh
forthommel: True! I will indeed make use of them in my next push.

monitors/rootmonitor/src/ROOTMonitor.cc
forthommel: Would it make sense to add a `eudaq::Status::State` accessor to the `eudaq::CommandReceiver` API, or possibly turn `m_status` into a protected member?

monitors/rootmonitor/include/eudaq/ROOTMonitorWindow.hh
forthommel: Addressed in 22654ad8880af49de4fe15f3f5c25eb51c486fb6.

forthommel:
@lhuth, @arummler, all,
Please let me/us know if you need any input from us to ease your review!

main/lib/ttree/src/TTreeEventConverter.cc
```diff
@@ -69,7 +69,9 @@ namespace eudaq{
     uint32_t id = d1->GetType();
     auto cvt = Factory<TTreeEventConverter>::MakeUnique(id);
     if(cvt){
-      return cvt->Converting(d1, d2, conf);
+      cvt->Converting(d1, d2, conf);
+      d2->Fill();
+      return true;
```
eyiliu: You might possibly the first user of the TTree converter.  
I'm not sure if the Fill() at line  67 and line 73 have race condition or not.
Do you have any test run with RootMonitor? Which kind of Event Data is feed into the TTreeConverter/RootMonitor?

monitors/rootmonitor/include/eudaq/ROOTMonitor.hh
```diff
@@ -0,0 +1,53 @@
...
+    void DoStopRun() override;
+    void DoTerminate() override;
+    void DoReset() override;
+    void DoReceive(eudaq::EventSP ev) override;
+
+    virtual void AtInitialisation() {}
+    virtual void AtConfiguration() {}
+    virtual void AtRunStart() {}
+    virtual void AtRunStop() {}
+    virtual void AtEventReception(eudaq::EventSP ev) = 0;
```
eyiliu: `AtEventReception(eudaq::EventSP ev) ` is required be to implemented in ROOT macro by end user. Am I wrong?

eyiliu:
@forthommel , it is nice implement of root monitor. Thank you very much

forthommel:
Thank you very much, @eyiliu !
Indeed, it is relatively tricky to judge without a proper example. After reading back the PR description I realised it is not so clear that the `TTree` producer fix is somewhat decorrelated from the ROOT monitor (and can actually be the topic of another non-overlapping PR)
My follow-up plan was to make use of this `TTreeEvent`-converted format (if implemented for the telescope/DUT streams) to fill directly a custom monitor view (without any need for an user-specific implementation of any monitor) pretty much like what is provided [here](https://github.com/eudaq/eudaq/blob/master/user/eudet/module/src/DirectSaveDataCollector.cc) or [there](https://github.com/eudaq/eudaq/blob/master/user/eudet/module/src/EventIDSyncDataCollector.cc) for `DataCollector`s.
I guess such an implementation is not too much of an effort though... And this might actually come in a follow-up PR.
For the time being, you may have a look at [our derived monitor](https://github.com/forthommel/eudaq/blob/cmstotem/user/cmstotem/module/src/SampicMonitor.cc) for our custom [data format](https://github.com/forthommel/eudaq/blob/cmstotem/user/cmstotem/extern/SampicEvent.hh). I will try to provide you a few screenshots/outputs (after some discussion with my colleagues of course...)

monitors/rootmonitor/include/eudaq/ROOTMonitor.hh
forthommel: You may find a (messy...) example (along with the `AtConfigure` part in https://github.com/forthommel/eudaq/blob/cmstotem/user/cmstotem/module/src/SampicMonitor.cc#L108-L186

main/lib/ttree/src/TTreeEventConverter.cc
forthommel: Indeed, this is what I understood from a previous conversation with @lhuth :-)
As told below, I will come up with a proper example after discussion with my colleagues on what may be released "to the world" from our previous test beam runs / lab fiddling.

monitors/rootmonitor/include/eudaq/ROOTMonitor.hh
forthommel: And indeed, I left this base class as a pure virtual object to ensure the monitor actually makes sense (and monitors something... :-) )

forthommel:
Digging in the issues, this PR might close/fix both #337 and #543.

forthommel:
Dear @eyiliu, @arummler, @lhuth 
Here are a few screenshots of our hardware-specific monitor (**not** included in this PR) "replaying" (unpacking and displaying) a preparation run taken in the lab a few weeks ago. As previously quoted, this correspond to [this derived object](https://github.com/forthommel/eudaq/blob/cmstotem/user/cmstotem/module/src/SampicMonitor.cc).

An empty view of the Sampic monitor (e.g. at startup):


A collection of per-channel monitors (double-click on any lhs "directory") while running/replaying a previously stored RAW file:


Peek at a single monitor for a specific channel:
Screenshot from 2019-10-17 08-23-55

Summary of same monitors for all channels in one board:
Screenshot from 2019-10-17 08-21-08

eyiliu:
Hi @forthommel ,
Thanks. Could you please quickly remind if there is any main difference between this PR code and the base ROOTMonitor in [cmstotem](https://github.com/forthommel/eudaq/tree/cmstotem) branch which you run with your raw data to generate above screenshots.

My main considering is your change in the  ```TTreeConverter``` by adding additional ```fill()```.  Presuming you made the test run with ```StdEventConverter``` rather than ``` TTreeConverter  ``` in your derived ```cmstotem``` monitor.  

Correct me if my understanding is wrong.

Yi

forthommel:
Hi @eyiliu,
As seen e.g. [here](https://github.com/forthommel/eudaq/compare/root-monitor...forthommel:cmstotem) there is no difference at all on this part of the code.

And indeed, I only realised now this extra `Fill()` in this other branch. You are right, the events unpacking is performed with a `StdEventConverter`-like operation (our data format is a derivative of `StandardEvent` which conveniently adds a RAW data-based  constructor performing the same operation than this converter).

Probably it would be safer for the time being to leave this `TTreeConverter` fix for a follow-up PR after a proper re-testing of everything (and possibly adding this `TTreeEvent`-based monitor alongside)?

forthommel:
@eyiliu I just dropped this `TTreeEventConverter` fix for the time being, leaving only the core feature of this PR.

monitors/rootmonitor/src/ROOTMonitor.cc
forthommel: @lhuth Any strong feeling on this?

lhuth: I personally do not have a strong opinion on that. However, @eyiliu is the core-expert and maybe has strong feelings on it.

forthommel: Thanks, @lhuth !

forthommel:
Hello @eyiliu, @lhuth, @arummler,
Any news on this review?

eyiliu:
Sorry @forthommel ,  please give me a few more days (2-3) to take care of the review with a run test in my local machine with my own dataset.

forthommel:
Thanks @eyiliu ! And apologies for my insistence...
Do not hesitate if you need any input from me (or sample datasets acquired with our subsystem).

forthommel:
Thanks a lot for these adaptations, @eyiliu ! Travis seems to be lagging behind to release its report, with one check failed due to a timeout.

arummler:
I re-triggered it (after the re-work of the CI the expensive update will no longer we necessary but it is not ready yet).

forthommel:
Thank you very much for your review, @eyiliu, @arummler, @lhuth !
I will work on the documentation part, and follow your suggestion of an example monitor (incl. a generic ROOT tree monitor for user-modules including a `TTreeEventConverter`).

Issue #592, opened 2019-11-01
Non-zero timestamp from triggered devices


mwilliams294:
When using the code fragment below with data from a triggered device (in our case mimosa26), it is expected that the timestamp from the plane should always be 0 :
```
auto plane = evt->GetPlane(static_cast<size_t>(plane_id));
for (unsigned int i=0; i < plane.HitPixels(); i++){
    if (plane.GetTimestamp(i) == 0){
        % do A %;
    } else {
        % do B %;
    }
```

However, it is observed that this value is 0 for some pixels in a plane but 1 for others, even though all pixels have the same trigger ID assigned to them.

lhuth:
Hi @mwilliams294,
Thanks for pointing that out. I could reproduce it. However, there seems to be no obvious reason why it happens. I'll dig deeper and keep everyone updated here.

lhuth:
I think I found it. See  #593. Could you check-out the branch and see if it is working for you now? 
And I think you wanted to use `plane.GetTimestamp(i)`, or?

mwilliams294:
Hi @lhuth,

#593 seems to have fixed the problem as I am only seeing `0` returned for the same set of mimosa26 data.

Thanks for the fix!

simonspa:
Fixed by #593

PR #593, opened 2019-11-04
Bugfix in the StdEvent

lhuth:
Digging into the issue #592 I found a bug in the NiRawDataConversion and the StandardEvent. The issue is present since the merge of PR #532. Fortunately, it is not affecting the pixel addresses, only the timestamps are wrongly/randomly assigned.
It would be nice if @mwilliams294 could test if her issues are fixed.

### DIFF
--- a/main/lib/core/src/StandardPlane.cc
+++ b/main/lib/core/src/StandardPlane.cc
@@ -50,9 +50,10 @@ namespace eudaq{
   }
 
   void StandardPlane::Print(std::ostream &os, size_t offset) const  {
-    os << std::string(offset, ' ') << m_id << ", " << m_type << ":" << m_sensor << ", " << m_xsize << "x" << m_ysize << "x" << m_pix.size()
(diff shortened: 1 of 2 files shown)


mwilliams294:
I tested #593 and it fixes #592

lhuth:
Thank you!

simonspa:
Wait, but this is not a fix but just changes the feature. Before it was a timestamp per plane - which you now changed to per pixel?

lhuth:
@simonspa, One could not get the timestamp per plane before, it was always a `plane.GetTimestamp(int)`.  If your intention was to have a per plane one, it should have been `plane.GetTimestamp()`. I assumed you want to have it per pixel and changed it accordingly. As it seems to be a working solution I would keep it this way if there are no objections.

simonspa:
Hm, let me look at my git log.

simonspa:
Nah, you were right. Sorry for the noise, good catch.

PR #597, opened 2019-12-05
Fix Installation of Target Files

simonspa:
Commit 526a258420b93534898db235d5b283cbbc0bed61 resulted in only `eudaqTargets.cmake` being copied to the install location, while `eudaqTargets-<BUILD_FLAVOR>.cmake` have been left behind. They are required because they contain the `IMPORTED_LOCATION` statement used to locate the libraries at link time.


simonspa:
Please merge and tag v2.4.3 because v2.4.2 fails to build due to this bug.

simonspa:
Yes, same here - I simply manually fixed it by copying over the other target files...

PR #603, opened 2020-01-13
euCliTluReader: fix bug to read fine timestamp and scalers

jekroege:
... from subevent tag correctly.

Discovered when looking at data taken at DESY with the AIDA TLU.


lhuth:
Nice catch, thanks @jekroege

PR #608, opened 2020-03-05
Solved error when using pedantic and overloaded-virtual flags

fravera:
Dear Eudaq developers,
as you may know, for the CMS Phase II SW we uses Eudaq to run at the Desy test beam.
Since last time we did a testbeam there, we raised the requirements in the C++ flags and now we are having problem in compiling the Ph2_ACF with your code.

The errors are due to the flag pedantic and overloaded-virtual
The first ones are caused by some ';' at the end of functions in Factory.hh that I removed.
The overloaded-virtual error caused is caused by "virtual void LogMessage::Print(std::ostream &) const" which hides "virtual void Status::Print(std::ostream &, size_t offset = 0) const"
I am not totally sure if the solution I propose in the PR is addressing the aim of this function, but a function name change would mean a much larger intervention that I'm not confident to do.

I would really appreciate if you can integrate these changes.

Many thanks in advance.
Cheers,
Fabio

### DIFF
--- a/main/lib/core/include/eudaq/Factory.hh
+++ b/main/lib/core/include/eudaq/Factory.hh
@@ -61,7 +61,7 @@ namespace eudaq{
       return nullptr;
     }
     return (it->second)(std::forward<ARGS>(args)...);
-  };
+  }
 
   template <typename BASE>
   template <typename ...ARGS>
@@ -90,7 +90,7 @@ namespace eudaq{
       init=false;
     }
     return m;
-  };
+  }
     
   template <typename BASE>
   template <typename DERIVED, typename... ARGS>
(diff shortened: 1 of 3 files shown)


lhuth:
Hi Fabio,

Thanks for the PR. Looks fine at first -  we'll test it and come back to you asap.

Cheers,
Lennart

eyiliu:
Dear @fravera ,  

Thank you to figure out the `;` in the template `factory`. We did not aware ot it at all since most compilers takes kind of _less restriction rule_. 
The correction of ` LogMessage::Print` with 2nd default offset value is definitely right.
Really appreciate your PR.

Yi

PR #609, opened 2020-04-08
Updates for Caribou, Timepix3 and AIDA TLU

simonspa:
This MR contains fixes for

* Caribou: fix correction of timestamp calculation for two devices
* Timepix3: fix some decoding issues, adjust SpidTrigger to be able to use it as auxiliary devcie for timestamping
* AIDA TLU: correctly read fine timestamps form subevent instead of the event

Please do not squash commits.


simonspa:
Thanks, @eyiliu ! Will this be 2.4.3?

PR #610, opened 2020-06-03
Scanning

schuetzepaul:
Attempt to solve two issues in the scan algorithm:

1. Runs can fail when the `RunControl` checks for the stop conditions before the producer used for event counting hasn't yet reset its event counter. Example: Usage of the TLU as event counter. Run is started, but TLU didn't react yet. Then the event count is still above the maximum number - the TLU (and others) are called to stop the run and the TLU goes into error state, because it's not yet running. Fixed by checking whether all producers are running before. Suggested by @lhuth .
2. The `start`, `stop` and `step` values are integer values casted to doubles, since the default value is an integer. I changed those to doubles to actually get floating point numbers out.

### DIFF
--- a/gui/src/euRun.cc
+++ b/gui/src/euRun.cc
@@ -651,6 +651,15 @@ void RunControlGUI::nextStep()
 
         std::this_thread::sleep_for(std::chrono::seconds(3));
         on_btnStart_clicked();
+
+        while(!allConnectionsInState(eudaq::Status::STATE_RUNNING)){
+            updateInfos();
+            QCoreApplication::processEvents();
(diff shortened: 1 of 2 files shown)


lhuth:
ah - that's a classical double work :)

schuetzepaul:
Indeed it is. Should I close mine since you're a bit ahead?

lhuth:
I cherry-pciked the `double` commit to also include your fix. 
If you don't mind, closing the PR would be bets :+1:

schuetzepaul:
Perfect, thanks!

PR #612, opened 2020-07-16
merge all latest (2019-2020) developments from Lycoris to central master branch

MengqingWu:
* all commits on lycoris-dev has **rebased** to the eudaq/master before this pull request.
Please let me know if anything needs to be changed.

* Changes applied in `eudaq/CMakeLists.txt`:
** lines 6-11: provide possibility to run EUDAQ2 inside anaconda environment -> for python producer user like lycoris.
** this does not matter for usual usage as it is under an `if` statement.

* Changes applied in `eudaq/main`:
** `main/exe/src/euCliReader.cxx` : read all timestamps in 64-bit instead of 32-bit uint.
** `main/lib/python/src/PybindProducer.cc` : add `DoStatus()` function for python producer - lycoris producer needs this.

* Changes applied in `eudaq/user/eudet`:
** `user/eudet/exe/src/euCliTeleReader.cxx ` : read all timestamps in 64-bit instead of 32-bit uint.

Other changes are applied inside `user/aidastrip` only, responsible by Lycoris team.


eyiliu:
@MengqingWu ,

> *Changes applied in eudaq/CMakeLists.txt:
> ** lines 6-11: provide possibility to run EUDAQ2 inside anaconda environment -> for python producer user like lycoris.
> ** this does not matter for usual usage as it is under an if statement.

The better method is set   `CMAKE_PREFIX_PATH`  and `CMAKE_LINK_DIRECTORIES_BEFORE`
in `cmake -DVar=`

We need to prevent the main CMakeList  to be too complex.

arummler:
I agree with Yi but is it not better to use FindPython3 and then to set e.g. Python3_ROOT_DIR on the command line?
https://gitlab.kitware.com/cmake/cmake/-/issues/19492
https://cmake.org/cmake/help/git-master/module/FindPython3.html#module:FindPython3

MengqingWu:
> @MengqingWu ,
> 
> > *Changes applied in eudaq/CMakeLists.txt:
> > ** lines 6-11: provide possibility to run EUDAQ2 inside anaconda environment -> for python producer user like lycoris.
> > ** this does not matter for usual usage as it is under an if statement.
> 
> The better method is set `CMAKE_PREFIX_PATH` and `CMAKE_LINK_DIRECTORIES_BEFORE`
> in `cmake -DVar=`
> 
> We need to prevent the main CMakeList to be too complex.

I agree, I will try this one and let you know once it is updated. Thanks for the suggestion.

MengqingWu:
> I agree with Yi but is it not better to use FindPython3 and then to set e.g. Python3_ROOT_DIR on the command line?
> https://gitlab.kitware.com/cmake/cmake/-/issues/19492
> https://cmake.org/cmake/help/git-master/module/FindPython3.html#module:FindPython3

not sure how the anaconda links to the findpython3?

user/aidastrip/module/src/kpixRawEvent2StdEventConverter.cc.runningMean
```diff
@@ -0,0 +1,493 @@
+/*
```
eyiliu, line 1: Not sure if you need this file. Looks like a backup file

user/aidastrip/CMakeLists.txt
```diff
@@ -1,35 +1,56 @@
...
+set(KPIX_LIBRARY kpix::kpix)
+
+find_package(ROOT QUIET)
+if(ROOT_FOUND)
+  list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
+endif(ROOT_FOUND)
+
 ##-- OPT: Compile AIDA strip?
-option(USER_AIDASTRIP_BUILD "build AIDA strip" OFF)
+cmake_dependent_option(USER_AIDASTRIP_BUILD "build AIDA strip" ON "LibXml2_FOUND;kpix_FOUND;ROOT_FOUND" OFF)
```
eyiliu, line 15: suggest to  to make this option default OFF and move to the top of this file.

eyiliu:
> > I agree with Yi but is it not better to use FindPython3 and then to set e.g. Python3_ROOT_DIR on the command line?
> > https://gitlab.kitware.com/cmake/cmake/-/issues/19492
> > https://cmake.org/cmake/help/git-master/module/FindPython3.html#module:FindPython3
> 
> not sure how the anaconda links to the findpython3?

Unfortunately,  we are not able to test under anaconda.  To support  anaconda env might not be the strong enought reason to change the main CMakeLists.
 
For you information,  EUDAQ2 is current using old style FindPythonLibs , instead of modern FindPython3.

lhuth:
@kraemeru: Last call before we will close this due to inactivity

lhuth:
this seems to be inactive for way too long. Please reopen in case....

PR #614, opened 2020-08-18
fix the installed cmake file path when installation is out of source tree

eyiliu:


CMakeLists.txt
```diff
@@ -139,16 +139,16 @@ if(DEFINED EUDAQ_TTREE_LIBRARY)
   list(APPEND CONF_EUDAQ_LIBRARIES eudaq_${EUDAQ_TTREE_LIBRARY})
 endif()
 
-set(CONF_INCLUDE_DIRS "\${EUDAQ_INSTALL_DIR}/include")
+set(CONF_INCLUDE_DIRS "\${EUDAQ_INSTALL_DIR}/include \${EUDAQ_INSTALL_DIR}/extern/include")
```
eyiliu, line 142: the external dependency, mainly lcio is installed at extern/include.   This path is reqiured when build LCIO converter by import eudaqTargets.cmake

CMakeLists.txt
```diff
@@ -157,16 +157,13 @@ write_basic_package_version_file(eudaqConfigVersion.cmake
 # Install the export set for use with the install-tree
 install(EXPORT eudaqTargets
   NAMESPACE eudaq::
-  DESTINATION ${PROJECT_BINARY_DIR}
+  DESTINATION cmake
   COMPONENT dev)
 
-file(GLOB TARGET_FILES "${PROJECT_BINARY_DIR}/eudaqTarget*.cmake")
```
eyiliu, line 163: main change here.   eudaqTargets goes into ${PROJECT_BINARY_DIR}  which is 'build' folder, instead of a installed folder

PR #615, opened 2020-08-18
build tlu dataconverter by default

eyiliu:
Fix.  when TLU code was moved folder usr/tlu, but  cmake option  was not adjusted.
TLU dataConverter will be built by this change. However,  TLU producer is not able to be built without extra dependency.


simonspa:
@arummler could you review this? I'm not familiar with EUDAQ's CI. Thanks!

lhuth:
@arummler  @eyiliu 
Just found some time to update the MRs. Is this one still relevant? If so could you have a look @arummler? If not can we close it?

lhuth:
closed due to inactivity

PR #616, opened 2020-08-18
lib/lcio: fix cmake for compile external project against eudaq and lcio

eyiliu:
Explained as title. 
External project was not able to build against eudaq when eudaq_lcio had been enabled.


eyiliu:
Dear @lhuth and @simonspa , could you please peer  review  and merge it? I'm not able to merger a MR from my own side.

lhuth:
seems like that one slipped through...

looks fine  - cannot test it though as I am not having lcio around...

lhuth (approved):
no obvious issues. could not test building it though...

PR #620, opened 2020-10-26
Add StdEvent Converter for CMS Phase2

gingercheng:


lhuth (approved):
Thanks for adding another converter :+1: 
Nothing to complain from my side

simonspa (approved):

PR #621, opened 2020-11-04
Build only converter

lhuth:
I've added a build option to only build a converter, that allows building the TLU-producer on servers wit afs access. 
` BUILD_USER_TLU_ONLY_CONVERTER` defaults to off and only effects the TLU. :+1:


simonspa:
Either this should be supported for all modules or only called `USER_LTU_BUILD_CONVERTER` or so. Right now it gives the wrong impression that once could build all converters with it.

lhuth:
That's why I said I can extend it to all modules if you agree with the approach

simonspa:
I don't think that's a viable option because in many cases the converter depends on DAQ code (Caribou, CMS, ...) so it will be Flickwerk in any case.

`BUILD_USER_ONLY_CONVERTER_FOR_THOSE_MODULES_THAT_SUPPORT_IT` :smirk:

lhuth:
Still, you could decide not to build the producers in these cases, right?

lhuth:
works on lxplus :)

simonspa (dismissed):

simonspa (approved):

PR #624, opened 2020-11-17
TLU converter: do not presume that we have a config object

simonspa:
StdEventMonitor does not provide a `EudaqConfigSPC` object but a `nullptr` which crashes the converter unless we account for it.

@jekroege @lhuth


lhuth (approved):

jekroege:
Thanks for the bug fix!

PR #625, opened 2020-11-17
Always update the config beofre sending it

lhuth:
@simonspa  found a bug that the configuration file is not updated if a single connection is (re)-configured. This MR fixes this. Now the config/ini is automatically updated each time before it is used


lhuth:
Anyone any comments on this topic?

simonspa:
Yes, my comment is: :heart_eyes:

simonspa (approved):

PR #627, opened 2020-11-18
refactored c++ standard handling

cburgard:
The different sections of eudaq have an inconsistent handling of c++ standards. Some CMakeLists.txt files of some components explicitly set a c++ standard flag, while the main CMakeLists.txt file of the package explicitly sets a slightly different one, depending on the system and build environment. This can cause the compilation to fail on some systems, notably with very new compilers (g++ >= 9).
I have tried to harmonize the c++ standard flag throughout the package, but of course, I don't know if this breaks compilation on relevant systems.
Please review!


arummler:
I don't think it is a good idea to check for compiler flags by hand (as you are not covering all compilers). Would e.g. an inbuilt mechanism like "target_compile_features" not be better?

simonspa:
One more thing to this: if we compile against other packages (e.g. ROOT) we also have to make sure to pick up their used standard because of ABI incompatibilites. See e.g. https://gitlab.cern.ch/corryvreckan/corryvreckan/-/blob/master/CMakeLists.txt#L131-142

cburgard:
> One more thing to this: if we compile against other packages (e.g. ROOT) we also have to make sure to pick up their used standard because of ABI incompatibilites. See e.g. https://gitlab.cern.ch/corryvreckan/corryvreckan/-/blob/master/CMakeLists.txt#L131-142

Yes, I actually tried this approach first, but I failed to find where to put this in eudaq, as it seems like there is no central place where ROOT is included, hence there is no central place to set the compiler flags.

What do you think is the best solution here? Search for ROOT optionally in the platform config and set the compiler flags accordingly if found?

cburgard:
> I don't think it is a good idea to check for compiler flags by hand (as you are not covering all compilers). Would e.g. an inbuilt mechanism like "target_compile_features" not be better?

Yes, probably. I just was a bit unsure whether you have any specifications w.r.t. which cmake version I'm allowed to require for this.

cburgard:
I now adapted the code from corryvreckan to check for ROOT and use the ROOT compiler flag if available. I'm still explicity using
```
check_cxx_compiler_flag(-std=c++17 SUPPORT_STD_CXX17)
check_cxx_compiler_flag(-std=c++14 SUPPORT_STD_CXX14)
```
though because I don't know if/how this could be done with the ```target_compile_features``` feature instead. Any suggestions?

arummler:
Appveyor is now failing because it is still running with root 5. I did my CI development on the v1.x-dev branch and did not port it yet. I can try to do it in the next days.

cburgard:
This is just to ping and see if there is a chance to get this merged anytime soon. Is there something I can do to speed up the process?

eyiliu:
Thanks to @cburgard   
I believe one of the origin principle is to have the eudaq core lib as independent as possible. The minimum C++ standard which the eudaq core can build, is C++11.  We most likely have C++14 (gcc4.8/9), but not all have C++17.

The issue the miss-match C++ standard is critical when building root related part of eudaq  (ROOT monitor, TTree lib).  If I understand correctly,  ROOT adapts the string_view from C++17,  and implement itself when C++17 is not available or not enabled. It impacts the RStringView.hxx header file which could be a generated header to match ROOT's C++17 build option
statically.  Therefore, including the ROOT header will enforce our code to use the same C++ standard version. It is really annoy.

For ROOT itself, the older version ROOT C++ 17 feature likely not enabled  even the compiler supports C++17 .  Therefore,  the check for compiler C++ capability might not sufficient.

As to current CMake file, it set the C++14 for global, and also C++14 for ROOT related code.  I can not remember why, but possibly when global set C++17,  the ROOT related code can keep to C++14.  It  helps for a few years when C++17 feature is not widely enabled in ROOT installation.

cburgard:
Hi Yi!
Thanks for reviewing this PR in so much detail. Allow me to reply to your concerns.
Yes, the issue comes from the string_view in C++17. I do not know why this happens, but I can see that when I compile eudaq on my laptop with VERBOSE=1, I see that both c++14 and c++17 are enabled at the same time, once via "std=c++17", once via "gnu++14" . This is clearly not intended. I do not know where exactly in the (quite complicated) CMakeLists.txt hierarchy of eudaq these two flags appear, but I found that it was probably the easiest (and safest) option to just set the standard globally once to whatever is the highest available one and avoid any future inconsistencies.
As Simon rightfully pointed out, there is also the concern that the standard needs to match what ROOT was compiled against, so I added a piece of logic with the intent of checking the flags used for ROOT *if ROOT is available*, and otherwise picking whatever is the highest available standard.
This should be the safest option available under all circumstances, as this does not require ROOT (flagged as optional), but only checks for its presence to deicde on the compiler flag, while also using whatever is the highest available standard that the user-supplied compiler can provide.
As far as I can tell, this should simplify everyones life. If you can think of a scenario where this approach fails, but the currently implemented approach would succeed, please let me know!

cburgard:
@arummler did you have any time to follow up?

arummler:
Sorry, I need a few more days.

cburgard:
@arummler any chance to look at this yet?

simonspa:
> I'm not against it if it makes the life easier.

Let's go then.

lhuth:
Does not work on my laptop:
```
-- Determined EUDAQ version v2.4.4-128-g7ed9b8ef
-- lib/lcio library is NOT to be built (EUDAQ_LIBRARY_BUILD_LCIO=OFF)
CMake Error at main/lib/ttree/CMakeLists.txt:44 (endif):
  Flow control statements are not properly nested.


-- Configuring incomplete, errors occurred!
```

lhuth:
Tested and functional in tb21/22

herkerta (approved):

PR #628, opened 2020-11-24
added policy to use new cmake policy to allow overriding options

cburgard:
When adding eudaq as a submodule to other repositories, it's currently impossible to override options from master CMakeLists.txt files.
This can be fixed by requesting to explicitly use the "new" policy CMP0077.


eyiliu (approved):

eyiliu:
Thanks to @cburgard

PR #629, opened 2020-12-14
Hotfix logic error in TLU cmake

lhuth:
There has been an error in the configuration logic for the tlu part: 
The producers have not been built in some configurations in which they were supposed to. Additionally, I made the output a bit more verbose.


lhuth:
Does anyone want to comment on this?

simonspa (approved):

PR #630, opened 2021-01-21
V1.x dev

sejlahadzic:


simonspa:
Hej @sejlahadzic 

currently you are reading in a file for every single event, just to know which sensor geometry you are using - if I interpret your code correctly. You can do this much simpler because Corryvreckan allows you to pass configuration parameters to the EUDAQ event converter. Have a look at the README in `user/caribou` and the code in `user/caribou/module/src/CLICTDConverterPlugin.cpp` for examples.

Please remove the `SET_CXX_STANDARD` from the merge request, there is an open ticket discussing how to deal with the issue: #627

sejlahadzic:
Hi @simonspa, thank you for your comment. When you say configuration file, do you mean Corryvreckan main configuration file or detector file, or you are referring to some other file?

simonspa:
Hi, Corry main config file, in the `[EventLoaderEUDAQ2]` section, check the documentation: https://gitlab.cern.ch/corryvreckan/corryvreckan/-/tree/master/src/modules/EventLoaderEUDAQ2#configuring-eudaq2-event-converters

sejlahadzic:
Hi @simonspa, I know it has been a very long time but I finally tried to get back to this. We are still using EventLoaderEUDAQ, so I first had to add this calibration file there. I tried doing it in a similar way to how it is done for EventLoaderEUDAQ2, but I believe there the shared pointer ConfigurationSPC is used (sorry if the terminology is wrong), which I think is not there in the eudaq1 Configuration class. Do you have any other suggestions on how to pass the calibration file from Corry to eudaq1?

simonspa:
Sorry, no, EUDAQ 1 is not maintained anymore for several years, and I have no idea how to implement something like this there.

sejlahadzic:
Thank you @simonspa. 
What should be adjusted if we want to change to EUDAQ2? I assume this plugin should be adjusted, but probably there are more things?

simonspa:
Actually, switching to EUDAQ2 is quite simple, which is why I don't fully understand why people still use EUDAQ1. :)

A few functions have to be renamed for the producer, and the structure of the decoder changes a little - but there are many examples in the `user/` folder, just check out one of them and model yours after them.

lhuth:
More than a year of silence. Closing it now

PR #632, opened 2021-03-04
Updating configruation for aida-tlu firmware 0x26 & adding pmt voltages to eudet/aida_tlu configs

lhuth:
We have updated the default firmware of the AIDA-TLUs at the DESY-II test beam. This comes with a few changes in the address tables that we now load as default.
In addition, the PMT voltages options are added to all config files.


simonspa:
Looks flawless to me - thanks for all the work! :tada:

simonspa (approved):

PR #634, opened 2021-05-05
Timepix3: fix nullptr access when called w/o configuration

simonspa:


lhuth:
All changes look reasonable and only affect the module for the TPX3 producer. Thanks for the update

lhuth (approved):

PR #639, opened 2021-08-10
Updates from latest tesbeam campaign, some cleanup and new toolz

superbello:
Updated all decoders, modularised a bit, added tools for re-synchronisation and simple occupancy checks.


superbello:
Moinsen - not sure there are many of us on Github. I did run this by B. Gallop who is the lead daq software developer for this and has made quite a few of the edits himself. Let me know if you want me to persuade someone to specifically look at this repo. (Noting here that people tend to scatter on public repository sites (bitbucket/github), whereas everyone in the field can get to gitlab at CERN, any reason the main eudaq repo wouldn't be moved to CERN gitlab?)

arturo-rodriguez35:
Hallo. I would only have one question for @superbello. From what I see, the total number of channels for the super plane is hardcoded for barrel modules. So while it is true that we can't use the super plane for endcap modules right now, I think we should make clear that this won't work for endcap modules, particularly for the R0. Do you think at some point it may be better to have two different converters? Maybe, there is a better solution to this directly from ITSDAQ.

superbello:
Hi @arturo-rodriguez35 - Whils the old convertor generates individual planes of the right size, it definitely doesn't work out of the box. I just pushed a changed superplane that covers the maximum size. As this size doesn't really affect anything, hits are stored as hits and sizes are not read by anyone afaik, this won't change the function as is. However, we should start looking at corry and how it "receives" these planes, as the current configurations hardwire a size in corry, that needs adapting depending on what is actually in the individual hits stored on the plane (the size of the plane does not play into that - funnily no checks on what hits get pushed into a plane...)

lhuth:
> Moinsen - not sure there are many of us on Github. I did run this by B. Gallop who is the lead daq software developer for this and has made quite a few of the edits himself. Let me know if you want me to persuade someone to specifically look at this repo. (Noting here that people tend to scatter on public repository sites (bitbucket/github), whereas everyone in the field can get to gitlab at CERN, any reason the main eudaq repo wouldn't be moved to CERN gitlab?)

Hi @superbello,
Not everyone is having a cern account and hence having write access to the repositories at `gitlab.cern.ch`.  The TB community has users not affiliated with cern or any LHC experiment. I personally do not have a strong opinion on how/where EUDAQ should be hosted, but do also not see a strong reason to transfer the repository.

superbello:
Hi @lhuth - Accepted, leaving it there.

superbello:
Hi @lhuth  (or others) - all settled with this last commit, I'll leave it there and would be happy if someone could merge.

lhuth (approved):
Since these are only changes on the itkstip side I'll approve and merge it

PR #642, opened 2021-09-14
Timepix3: some improvements

simonspa:
* Silence known bogus `0x6000` error on initialize
* Improve run loop
* Reduce timeout for SPIDR buffer read to 300ms in the hope that it would create smaller chunks

### DIFF
--- a/user/timepix3/module/src/Timepix3Producer.cxx
+++ b/user/timepix3/module/src/Timepix3Producer.cxx
@@ -917,41 +917,36 @@ void Timepix3Producer::RunLoop() {
   // create a thread to periodically ask for full timestamps, temperatures and others
   ts_thread = std::thread(&Timepix3Producer::timestamp_thread, this);
 
-  //int device_nr = 0;
-  //std::map<int, int> header_counter;
+  std::map<int, int> header_counter;
 
   // set sampling parameters
   spidrdaq->setSampleAll( true );
   // Sample pixel data
   spidrdaq->setSampling( true );
-  //spidrdaq->startRecording( "test.dat", 123, "This is test data" );
 
   while(m_running) {
 
-    bool next_sample = true;
     if(spidrdaq->bufferFullOccurred()) {
       EUDAQ_WARN("Buffer overflow");
     }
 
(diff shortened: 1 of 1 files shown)


user/timepix3/module/src/Timepix3Producer.cxx
```diff
@@ -1000,7 +991,7 @@ void Timepix3Producer::RunLoop() {
   ts_thread.join();
   ts_thread.~thread();
   // show header statitstics
-  //std::cout << "Headers: " << listVector(header_counter) << endl;
+  std::cout << "Headers: " << listVector(header_counter) << endl;
```
lhuth, line 994: This could be also a log message of some level?

user/timepix3/module/src/Timepix3Producer.cxx
```diff
@@ -917,41 +917,36 @@ void Timepix3Producer::RunLoop() {
...
+      // ...until the sample buffer is empty
+      while(uint64_t data = spidrdaq->nextPacket()) {
 
-        uint64_t header = (data & 0xF000000000000000) >> 60;
-        //header_counter[header]++;
+        // Break if we're supposed to stop:
+        if(!m_running) break;
 
-        //std::cout << "Data is: 0x" << to_hex_string(data) << std::dec << std::endl;
+        uint64_t header = (data & 0xF000000000000000) >> 60;
```
lhuth, line 948: As far as I know the `& 0xF000000000000000` is not needed if one shifts by 60?

user/timepix3/module/src/Timepix3Producer.cxx
simonspa: yeah, but it's of limited use and will probably just confuse users. I want to look at it for a while and then either promote it or remove it.

user/timepix3/module/src/Timepix3Producer.cxx
simonspa: I didn't touch any of the data bits, so I'd prefer not to do that in this MR.

lhuth (approved):

PR #643, opened 2021-09-20
ITS3 code

MagnusMager:
This contains the ITS3 development squashed into a few commits. This is from a fork from some year ago... feel free to cherry-pick.
It contains:
- updated python modules (including fixing some threading problems)
- CLI automated run control + some racetime condition fixes that appeared
- ITS3 specific producers


lhuth:
Hi @MagnusMager,

thanks for the producer update. Since this PR is rather big, reviewing will take a bit.
The changes to the c++ core part are all fine, the ITS3 specific cannot be fully evaluated by us, so we will accept them.
The last missing part are the python binding changes, that we will look at next.

MagnusMager:
Yes, no worries! We will try to keep more aligned in future.
The ITS3-specific producers will also rely on other python libraries (and eventually hardware) that we do not intend to include in eudaq. Of more general interest could be the the python run control (which is text mode and can run nicely in a tmux, ssh etc.), the python data collector, and the python data reader/dumper.

PR #647, opened 2021-11-30
Caribou -- Tangerine at DESY II Beamline 21

ffeindt:
- Introduced converter for peary device DSO9254A
- euCliReader now accepts configuration files and passes them to the eudaq::StdEventConverter


main/exe/src/euCliReader.cxx
```diff
@@ -78,7 +97,8 @@ int main(int /*argc*/, const char **argv) {
       ev->Print(std::cout);
       if(stdev_v){
         auto evstd = eudaq::StandardEvent::MakeShared();
-        eudaq::StdEventConverter::Convert(ev, evstd, nullptr);
+        //eudaq::StdEventConverter::Convert(ev, evstd, nullptr);
```
simonspa, line 100: Remove commented line, fix indentation?

main/exe/src/euCliReader.cxx
```diff
@@ -34,6 +34,25 @@ int main(int /*argc*/, const char **argv) {
   uint32_t timestamph_v = timestamph.Value();
   bool not_all_zero = eventl_v||eventh_v||triggerl_v||triggerh_v||timestampl_v||timestamph_v;
 
+
+  // load configuration
+    // empty configuration object, prevents crash if no config file given
+  eudaq::Configuration eu_cfg( "", "" );
```
simonspa, line 40: Couldn't you directly make this a `shared_ptr` instead of copying the thing later on?

```cpp
auto eu_cfg = std::make_shared<eudaq::Configuration>("","");
eu_cfg->Load(/*...*/);
```

main/exe/src/euCliReader.cxx
```diff
@@ -34,6 +34,25 @@ int main(int /*argc*/, const char **argv) {
...
+  }
+  // prevent warning if no config file name given
+  else if ( file_conf.Value().empty() ){}
+  // but warn if attempt failed
+  else{ 
+    std::cout << "WARNING, config file '" << file_conf.Value() << "' not found!" << std::endl;
+  }    
+  // build shared pointer
+  const eudaq::Configuration const_eu_cfg = eu_cfg;
+  eudaq::ConfigurationSPC config_spc = std::make_shared<const eudaq::Configuration>(eu_cfg);
```
simonspa, line 53: Ah, I see, it needs to be `const`. Hm. Well. Then keep it? :)

main/lib/core/src/StdEventConverter.cc
```diff
@@ -7,7 +7,6 @@ namespace eudaq{
   Factory<StdEventConverter>::Instance<>();
   
   bool StdEventConverter::Convert(EventSPC d1, StdEventSP d2, ConfigurationSPC conf){
-
```
simonspa, line 10: unrelated change?

user/caribou/CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ get_filename_component(EUDAQ_USERNAME ${CMAKE_CURRENT_LIST_DIR} NAME)
 set(EUDAQ_MODULE module_${EUDAQ_USERNAME})
 set(USER_HARDWARE_LIBRARY ${EUDAQ_USERNAME}_static)
 
+# DSO9254A converter needs root
```
simonspa, line 5: An alternative to *always* requiring ROOT would be to only build the DSO converter if ROOT is found.
But I actually think it's fine this way, otherwise things get overly complicated.

user/caribou/module/src/CaribouProducer.cxx
```diff
@@ -242,7 +242,7 @@ void CaribouProducer::RunLoop() {
         data_buffer.clear();
       }
 
-      LOG_PROGRESS(STATUS, "status") << "Frame " << m_ev;
+      //LOG_PROGRESS(STATUS, "status") << "Frame " << m_ev;
```
simonspa, line 245: Unrelated change

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+#include "time.h"
+
+using namespace eudaq;
+
+namespace{
+  auto dummy0 = eudaq::Factory<eudaq::StdEventConverter>::
+  Register<DSO9254AEvent2StdEventConverter>(DSO9254AEvent2StdEventConverter::m_id_factory);
+}
+
+uint64_t timeConverter( std::string date, std::string time );
```
simonspa, line 15: I think this should be a `static` member function of `DSO9254AEvent2StdEventConverter`.

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+
+  auto ev = std::dynamic_pointer_cast<const eudaq::RawEvent>(d1);
+  // No event
+  if(!ev) {
+    return false;
+  }
+
+
+  // load parameters from config file
+  double pedStartTime = conf->Get("pedStartTime", 0 ); // integration windows in [ns]
```
simonspa, line 28: You need to explicitly check here if the `config` is not `nullptr` otherwise this converter segfaults when called without a valid `config` object.

simonspa: Furthermore, these values are parsed agan and again for *every single event*. Maybe consider making them static member variables and to initialize them only once (have a look at the other Caribou converters that do something like this already).

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+  double pedEndTime   = conf->Get("pedEndTime"  , 0 );
+  double ampStartTime = conf->Get("ampStartTime", 0 );
+  double ampEndTime   = conf->Get("ampEndTime"  , 0 );
+  bool generateRoot   = conf->Get("generateRoot", 0 ); 
+  std::cout << "Loaded parameters from configuration file." << std::endl
```
simonspa, line 33: Please do not use `std::cout` because then there is no way of controlling the verbosity level. This seems to be something appropriate for `EUDAQ_DEBUG`?

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+    histoFile = new TFile( "waveforms.root", "UPDATE" ); // UPDATE?
+    if(!histoFile) {
+      LOG(ERROR) << "ERROR: " << histoFile->GetName() << " can not be opened";
+      return false;
+    }
+  }
+						       
+  
+  // Retrieve data from event
+  if(ev->NumBlocks() == 1) {
```
simonspa, line 71: you check for number of blocks == 1 but don't do anything if they're *not* 1. Is this intended? Then data is simply skipped without notification or anything?

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+
+    
+    // all four scope channels in one data block
+    auto datablock = ev->GetBlock(0);
+    LOG(DEBUG) << "DSO9254A frame with "<< datablock.size()<<" entries";
+
+    
+    // Calulate positions and length of data blocks:
+    // FIXME FIXME FIXME by Simon: this is prone to break since you are selecting bits from a 64bit
```
simonspa, line 80: This we should indeed discuss. :)

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+    // loop 4 channels
+    for( int nch = 0; nch < 4; nch++ ){      
+
+      std::cout << "Reading channel " << nch << std::endl;
+
+      // Obtaining Data Stucture
+      block_words = rawdata[block_posit + 0];
+      pream_words = rawdata[block_posit + 1];
+      chann_words = rawdata[block_posit + 2 + pream_words];
+      std::cout << "  " << block_posit << " current position in block\n";
```
simonspa, line 105: See above about `std::cout`

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+      std::cout << "  " << block_words << " words per segment\n";
+      std::cout << "  " << pream_words << " words per preamble\n";
+      std::cout << "  " << chann_words << " words per channel data\n";
+      std::cout << "  " << rawdata.size() << " size of data rawdata\n";
+           
+
+      // read preamble:
+      std::string preamble = "";
+      for(int i=block_posit + 2;
```
simonspa, line 114: Do we really need to ad this byte-by-byte? `std::string` has great constructors which should also take a memory position and a length or something...

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+      dx.push_back( stod( vals[4]) * 1e9 );
+      x0.push_back( stod( vals[5]) * 1e9 );
+      dy.push_back( stod( vals[7]) );
+      y0.push_back( stod( vals[8]) );
+      if( vals.size() == 25 ) {// this is segmented mode, possibly more than one waveform in block
+	sgmnt_count = stoi( vals[24] );
+      }
+      int points_per_words = np.at(nch)/(chann_words/sgmnt_count);
+      if( np.at(nch)%chann_words/sgmnt_count ) LOG(WARNING) << "incomplete waveform";
+      /*
```
simonspa, line 145: Please remove commented code - unless you need it, then uncomment it! ;)

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+      std::vector<double> amp;
+      std::vector<std::vector<double>> waves;
+      
+      
+      for( int s=0; s<sgmnt_count; s++){ // loop semgents
+
+     	// container for one waveform
+	std::vector<double> wave;
+	
+	// prepare histogram with corresponding binning
```
simonspa, line 163: Shouldn't these only be prepared if `generateRoot` is true? Now we create them even if we don't want to write out to ROOT file...

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+	hist->GetXaxis()->SetTitle("time [ns]");
+	hist->GetYaxis()->SetTitle("signal [V]");
+      
+	// read channel data
+	std::vector<int16_t> words;
+	words.resize(points_per_words); // rawdata contains 8 byte words, scope sends 2 byte words 
+	int16_t wfi = 0;
+	for(int i=block_posit + 3 + pream_words + (s+0)*chann_words/sgmnt_count;
+	        i<block_posit + 3 + pream_words + (s+1)*chann_words/sgmnt_count;
+	      //i<block_posit + 3 + pream_words+chann_words;
```
simonspa, line 182: :thinking:

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+	      amps.at(c).at(s) = wavess.at(c).at(s).at(p) - peds.at(c).at(s);
+	  }
+	}
+
+	//}catch(...){} 
+	
+      } // points
+      
+      
+      // check:
```
simonspa, line 267: See above about commented code

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+  if( time.size() > 0 ){ // only if there are waveform data
+  
+
+    // re-iterate waveforms
+    // wavess.at(channel).at(segment).at(point)
+    // ped   .at(channel).at(segment)
+    for( int s = 0; s<peds.at(0).size(); s++ ){
+      for( int p = 0; p<time.size(); p++ ) {      	
+	
+	//try{ // hotfix in case off incomplete waveforms
```
simonspa, line 246: Is this still necessary? Do we see incomplete waveforms?

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+		  << "         ampli   " << amps.at(c).at(s) << std::endl
+		  << "         timestp " << timestamp << std::endl;
+      }
+      */
+  
+    } // segments
+    
+    
+    // Create a StandardPlane representing one sensor plane
+    // FIXME For now pushing just the first segment... need to decide how to treat several of them
```
simonspa, line 283: That's a good question - normally we have these separated in subevents - but at the stare of RawEvents from eudaq, not after decoding... :thinking:

user/caribou/module/src/DSO9254AEventConverter.cc
```diff
@@ -0,0 +1,377 @@
...
+  else{
+    
+    // close rootfile anyway
+    if( generateRoot ){
+      histoFile->Close();
+      std::cout << "Histograms written to " << histoFile->GetName() << std::endl;
+    }
+
+    
+    std::cout << "Warning: No scope data in event " << ev->GetEventN() << std::endl;
```
simonspa, line 312: Here and above are a few more `std::cout`s

simonspa (changes requested):

user/caribou/module/src/DSO9254AEventConverter.cc
simonspa: https://gitlab.cern.ch/corryvreckan/corryvreckan/-/merge_requests/489

main/exe/src/euCliReader.cxx
ffeindt: according to you comment below this is fine

user/caribou/module/src/DSO9254AEventConverter.cc
ffeindt: i think not, i removed that

user/caribou/module/src/DSO9254AEventConverter.cc
ffeindt: > You need to explicitly check here if the `config` is not `nullptr` otherwise this converter segfaults when called without a valid `config` object.

Well `corry` aborts anyway if no config file is defined. `euCliReader` still starts but falls back to the default values (0 in all cases), so it does not produce a root file. The same is true for corry if the keys are not defined in the config file. Still I think think this behavior is not ideal. I will implement some sort of error message.

ffeindt: > Furthermore, these values are parsed agan and again for _every single event_. Maybe consider making them static member variables and to initialize them only once (have a look at the other Caribou converters that do something like this already).

Implemented. Although I don't get why these should be static.

user/caribou/module/src/DSO9254AEventConverter.cc
ffeindt: i think this is related to the memcpy, which we want to discuss any way

user/caribou/module/src/DSO9254AEventConverter.cc
ffeindt: yes

user/caribou/module/src/DSO9254AEventConverter.cc
simonspa: The way these decoders are implemented in EUDAQ2, there is no guarantee that the *same decoder object* will be used for consecutive events. It might as well be that a new object is created. In that case, suddenly all your variables are uninitialized instead of having the previously-assigned values from the configuration.

By making them static, the variable's scope is global and outside of the decoder objects. This means, any instance of this decoder will use the same values that you have assigned.

It's crooked but that's what the design of the decoder interface requires... :/

Does this make sense?

Issue #649, opened 2022-01-28
Problem when enabling building option: "EUDAQ BUILD STDEVENT MONITOR=ON" during installation


km-b214:
Hello, 

I have a problem during "make install" after I enable the building option: EUDAQ BUILD STDEVENT MONITOR=ON. Running "make install" with default setting directed from the manual is fine. Here is the error message: 

![image](https://user-images.githubusercontent.com/81219506/151484707-ee271d5b-391c-48db-9b97-991a37cb1711.png)

I wonder if anyone could help me with this. 

Thank you,

lhuth:
Hi @km-b214 ,

We have not observed this error yet. It stems from root. Which exact version of root are you using?
Cheers,

@lhuth

km-b214:
Hello @lhuth, 

The version of root installed is 6.24/02.  

thanks, 
@km-b214

km-b214:
I updated root to root 6.24/06 but the error message is still the same. 

@km-b214

lhuth:
We have only tested with versions up to 6.22 - I'll give it a try and come back to you
Cheers

lhuth:
Hi, I could reproduce the issue, which is due to a change in `TDirectory` from root 6.22 to 6.24. Pull request is open

lhuth:
Hi @km-b214,

everything should be working now in the master. Could you test it please? If everything is fine we can close this issue

km-b214:
Hi @lhuth,

Yes, it's working now. Thank you much. You may close this issue. 

@km-b214

PR #650, opened 2022-01-28
updating to new root version

lhuth:
Fix for issue #649 
to be tested on root 6.22 @herkerta

### DIFF
--- a/monitors/onlinemon/src/EUDAQMonitorCollection.cc
+++ b/monitors/onlinemon/src/EUDAQMonitorCollection.cc
@@ -32,7 +32,7 @@ void EUDAQMonitorCollection::Write(TFile *file) {
     std::cerr << "EUDAQMonitorCollection::Write File pointer is NULL" << endl;
     exit(-1);
   }
-  if (gDirectory != NULL) // check if this pointer exists
+  if (gDirectory) // check if this pointer exists
   {
     gDirectory->mkdir("EUDAQMonitor");
     gDirectory->cd("EUDAQMonitor");
--- a/monitors/onlinemon/src/MonitorPerformanceCollection.cc
+++ b/monitors/onlinemon/src/MonitorPerformanceCollection.cc
@@ -25,7 +25,7 @@ void MonitorPerformanceCollection::Write(TFile *file) {
     cout << "MonitorPerformanceCollection::Write File pointer is NULL" << endl;
     exit(-1);
   }
-  if (gDirectory != NULL) // check if this pointer exists
+  if (gDirectory) // check if this pointer exists
   {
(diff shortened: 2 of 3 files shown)


lhuth:
Tested with root 6.10 - good to go

monitors/onlinemon/src/EUDAQMonitorCollection.cc
```diff
@@ -32,7 +32,7 @@ void EUDAQMonitorCollection::Write(TFile *file) {
     std::cerr << "EUDAQMonitorCollection::Write File pointer is NULL" << endl;
     exit(-1);
   }
-  if (gDirectory != NULL) // check if this pointer exists
+  if (gDirectory) // check if this pointer exists
```
simonspa, line 35: `!= nullptr`?

monitors/onlinemon/src/MonitorPerformanceCollection.cc
```diff
@@ -25,7 +25,7 @@ void MonitorPerformanceCollection::Write(TFile *file) {
     cout << "MonitorPerformanceCollection::Write File pointer is NULL" << endl;
     exit(-1);
   }
-  if (gDirectory != NULL) // check if this pointer exists
+  if (gDirectory) // check if this pointer exists
```
simonspa, line 28: same

monitors/onlinemon/src/EUDAQMonitorCollection.cc
lhuth: Nope, not possible since root 6.24.

herkerta (approved):

monitors/onlinemon/src/EUDAQMonitorCollection.cc
simonspa: because it's not a pointer?

simonspa: Looking at the redefinition, a

```cpp
gDirectory() != nullptr
```

should work

lhuth: They have wrapped this into some construction to be threadsave, which overrides the operator != twice. 
TDirectory* dir = nullptr
If(gDirectory!=dir)

Would work, but they have a direct check implemented now 🤷

simonspa: Whatever... :)

simonspa (approved):

monitors/onlinemon/src/EUDAQMonitorCollection.cc
lhuth: It works....

PR #651, opened 2022-02-04
Move to C++17

simonspa:


lhuth (approved):

lhuth:
We should at some point make this adjustable. And just check that is not causing any trouble on the tb machines @herkerta

simonspa:
> We should at some point make this adjustable. And just check that is not causing any trouble on the tb machines @herkerta

That's done in https://github.com/eudaq/eudaq/pull/627

lhuth:
we will tackle #627 again

PR #652, opened 2022-02-04
Add Support for Caribou AD9249 65MHz ADC

simonspa:


lhuth (dismissed):
Hi @simonspa,

all fine for me. Concerning the changes in the ITS3 Converter. They should be transparent to users without any configuration file usage?

simonspa:
@lhuth ready for review, I fixed the issue with ALPIDE, now no config is required, old behavior is restored.

simonspa:
@lhuth ping please review and merge.

user/tlu/module/src/TluRawEvent2StdEventConverter.cc
```diff
@@ -126,11 +129,24 @@ bool TluRawEvent2StdEventConverter::Converting(eudaq::EventSPC d1, eudaq::Standa
   // with combined fineTS: replace the 3lsb of the coarse timestamp with the 3msb of the finets_avg:
   auto ts = static_cast<uint64_t>((25. / 32. * (((coarse_ts << 5) & 0xFFFFFFFFFFFFFF00) + (finets_avg & 0xFF))) * 1000.);
 
-  // Set times for StdEvent in picoseconds (timestamps provided in nanoseconds):
-  d2->SetTimeBegin(ts);
-  d2->SetTimeEnd(d1->GetTimestampEnd() * 1000);
-  d2->SetTimestamp(ts, d1->GetTimestampEnd(), d1->IsFlagTimestamp());
+  // store time of the run start
+  if(m_runStartTime < 0){
+    m_runStartTime = ts; // just use one of them for now
```
lhuth, line 134: Why would one set the run start to the first trigger time stamp? This causes trouble for any device that takes a t_0. This has to be zero if anything? As far as I see, that parameter is initialised with -1 and cannot be configured 🤷

user/tlu/module/src/TluRawEvent2StdEventConverter.cc
```diff
@@ -126,11 +129,24 @@ bool TluRawEvent2StdEventConverter::Converting(eudaq::EventSPC d1, eudaq::Standa
...
+  // Set times for StdEvent in picoseconds (timestamps provided in nanoseconds):
+  d2->SetTimeBegin(ts - m_runStartTime);
+  d2->SetTimeEnd(d1->GetTimestampEnd() * 1000 - m_runStartTime);
+  d2->SetTimestamp(ts - m_runStartTime, d1->GetTimestampEnd(), d1->IsFlagTimestamp());
+
+  // print time stamps to have a look
+  /*
+  std::cout << "TIMEINFO_TLU: itrg iev framestart frameend " << d1->GetTriggerN()
+	    << " " << d1->GetEventN()
+    	    << " " << ts
```
lhuth, line 146: I thought this was removed?

lhuth (changes requested):

user/tlu/module/src/TluRawEvent2StdEventConverter.cc
simonspa: it was two months ago, but we have apparently continued dev and it resurfaced. :) will fix

simonspa:
Both were only a rebase-gone-wrong, fixed now.

lhuth (approved):
:rocket:

PR #654, opened 2022-04-05
improved string handling

cburgard:
The current implementation of the config handler relies heavily on exceptions being thrown and caught.
This is inefficient and a pain when debugging.
This PR fixes that by switching to a non-exception way of handling configuration keys.


lhuth (approved):
completely fine with me - same as for the other. I just would like to let the code run once in the test beam just be sure :)

PR #656, opened 2022-04-05
Added dictionaries

cburgard:
eudaq does not generate any dictionaries. This makes it hard to use in interactive shells or with python.
This PR sets out to at least generate dictionaries for the exception classes, so that exceptions thrown within eudaq can be processed sensibly in python scripts.
Also, the implementation of the Exception classes is cleaned up a bit.

### DIFF
--- a/main/lib/core/CMakeLists.txt
+++ b/main/lib/core/CMakeLists.txt
@@ -7,11 +7,30 @@ set(EUDAQ_INCLUDE_DIRS
   ${CMAKE_CURRENT_LIST_DIR}/include/eudaq
   CACHE INTERNAL "directory of eudaq include")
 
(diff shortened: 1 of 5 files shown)


cburgard:
> This one seems fine. I just like to check it running on one of the telescopes

you mean live (as in, the online-monitor)?
or do you mean as a library for corry? The latter one works.

lhuth:
> > This one seems fine. I just like to check it running on one of the telescopes
> 
> you mean live (as in, the online-monitor)? or do you mean as a library for corry? The latter one works.

like in real at the tb - we had once an unexpected surprise running there, which I just never want to have again :)

herkerta (approved):

PR #661, opened 2022-07-08
Update build option for MacOS

BongHwi:
The current build option generates the `dylib` file based on Apple's default configuration.
To import the `pyeudaq` library to MacOS system, I forced to generate the `so` file.


lhuth (approved):
thanks @BongHwi

PR #662, opened 2022-07-11
fixing issue with newly introduced floAT GET() in config

lhuth:
Thanks to H. Augustin & U. Kraemer for spotting


simonspa:
An explanation *what* the issue was would be nice.

lhuth:
> An explanation _what_ the issue was would be nice.

Did not compile, since I overlooked an error in the implementation. Now I adjusted the style of getting it to all other routines used.

simonspa (approved):

arummler:
I really should take the time to finally fix the CI to avoid such incidents.

PR #664, opened 2022-07-20
StandardPlane: add possibility for waveforms

simonspa:
as title says.

I will improve the interface still a bit , but basic functionality is there.


simonspa:
All done.

PR #665, opened 2022-07-24
FIX: FSM does not allow connection in CONFIGURED state

arummler:
According to documentation the FSM should allow connections in CONFIGURED state. This was disabled both when starting a run (correct) and when configuring (wrong). Removed the second occurrence.


lhuth (commented):
Hi Andre,

To be honest. I've never checked this in the manual. However, I think this makes a lot of sense this way. @herkerta any objections?

herkerta (approved):

PR #674, opened 2022-08-09
Waveforms: SetupResults

simonspa:
This implements the weird CDS and ACCUMULATE stuff for waveforms, a bit hackily because we don't subtract the frames for CDS actually.


simonspa:
(cc) @lhuth

lhuth (approved):

PR #663, opened 2022-07-12
Draft: modernizing and fixing CI

arummler:
- changed ROOT C++ version recognition which did not work on Windows (: instead of =)
- removed travis which basically became useless for unpaid plans one year ago
- replacing travis with appveyor; using restructured config file which introduced first in the v1x-dev branch
- ... ongoing work ... will be squashed when ready ... do not merge yet

PR #676, opened 2022-08-10
Adding a minimal GitHub CI to ensure, that the converters do not cause a segfault

lhuth:
This PR is adding a (very limited and minimal) CI to avoid segfaults/breaking changes in the Event format with the following features:

* Compilation on latest ubuntu
* A single test: test_mimosa_tlu_io to check if data can be read without a segfault

Note that this can (and should be) extended.  As far as I understand the GitHub regulations, we have unlimited time to run tests while the repository is public. 

The test will only run after the merge with the current settings... Should we enable it for each commit? 🤔


arummler:
We had up to now a combination of travis and appveyor which I had mostly implemented. Travis became basically non-free and was removed. There was one thing broken which fixed in PR #663 with the goal to do linux and mac also on appveyor. But we can migrate to github actions if you prefer. Probably easier than a thrid party provider although I would need to look into the capabilities. I had tried to cover also all the user specific requirements as far as feasible. Would require porting if we should keep that up.

lhuth:
Hi @arummler,

I do not have hard feelings about which CI to use - I just went for the most straightforward one to have at least something running.
The nice part about  GiHutb is - afaik - that it comes for free and is likely to stay. 
I am not really convinced to test user code in the CI - this becomes messy pretty quickly (e.g. we would need to build peary for the caribou part,...) Windows and macOS are both also supported in the GitHub CI... 

You are the expert if it comes to porting the existing checks. For now, we can also have the two in parallel.

#663 has by the way a draft label 😏

lhuth:
ping @arummler

simonspa (approved):

PR #677, opened 2022-08-14
Logging improvement in example python producer & DC

msuljic:


simonspa (approved):
Thanks, nice addition to the example

PR #679, opened 2022-08-18
Logging related minor improvements

msuljic:


simonspa (approved):
Looking good, thanks for the fixes and the added flexibility in the instantiation order of the log collector.

PR #685, opened 2022-09-16
Multi variable scan

kraemeru:
Added a new variable "scanParallelTo" to the scanHelper class and .scan files to allow the change of multiple variables of the same or separate producers in a single scan step at the same time.

If this variable is not specified in the .scan file it simply behaves like the old scan to ensure it is backwards compatible.


gui/src/scanHelper.cc
```diff
@@ -73,42 +92,103 @@ bool Scan::setupScan(std::string globalConfFile, std::string scanConfFile) {
     return true;
 }
 
-void Scan::createConfigs(unsigned condition, eudaq::ConfigurationSP conf,std::vector<ScanSection> sec) {
-    if(condition == sec.size())
+//void Scan::createConfigs(unsigned condition, eudaq::ConfigurationSP conf,std::vector<ScanSection> sec) {
+//    if(condition == sec.size())
```
lhuth, line 96: Commented code can and should be removed

user/example/misc/scan/ExampleScan.scan
```diff
@@ -1,5 +1,5 @@
 [global]
-repeatScans = 1
+repeatScans = 0
```
lhuth, line 2: revert changes that are not relevant for the PR, like this one :)

user/example/misc/scan/ExampleScan.scan
```diff
@@ -16,7 +16,7 @@ name = Producer.my_pd0
 parameter = EX0_PLANE_ID
 
 [4]
-nested = 1
+#nested = 1
```
lhuth, line 19: or this :)

user/example/misc/example_startrun.sh
```diff
@@ -5,8 +5,8 @@ $BINPATH/euRun -n Ex0RunControl &
 sleep 1
 $BINPATH/euLog &
 sleep 1
-$BINPATH/euCliMonitor -n Ex0Monitor -t my_mon & 
-$BINPATH/euCliCollector -n Ex0TgDataCollector -t my_dc &
+#$BINPATH/euCliMonitor -n Ex0Monitor -t my_mon & 
+#$BINPATH/euCliCollector -n Ex0TgDataCollector -t my_dc &
```
lhuth, line 9: Removing this makes the script obsolete. Any reason why this is done?

user/example/misc/scan/ExampleScan2.scan
```diff
@@ -0,0 +1,40 @@
+[global]
+repeatScans = 0
+allowNested = 1 # overrides local nested arguments
+#configPrefix = "path/to/folder/scanned" // optional
+timeBasedScan = 1 # 1 = true :)
+timePerStep = 10 #in seconds for now
+nEventsPerStep   = 200
+
+[0]
+scanParallelTo = 0
```
lhuth, line 10: Why would I like to scan that in parallel to itself? Looks odd to me. What if it is not defined at all?

gui/include/scanHelper.hh
```diff
@@ -126,6 +128,15 @@ private:
      * @param sec Scan section beeing read
      */
     void createConfigs(unsigned condition, eudaq::ConfigurationSP conf, std::vector<ScanSection> sec);
```
lhuth, line 130: If I am not blind you have removed the function in the cpp. Why is it still here? :)

lhuth (changes requested):
Hi @kraemeru,

thanks for setting this up. Overall a nice feature. A few more or less picky comments are below.

user/example/misc/scan/ExampleScan2.scan
kraemeru: I am not sure what you mean by "that in parallel to itself" if it is about the fact that it affects the same producer this is simply to show that you can also change two variables of the same producer at the same time.

If scanParallelTo is not defined it simply counts each [X] variable as its own unit to ensure backwards compatibility with old scan files.

user/example/misc/example_startrun.sh
```diff
@@ -5,11 +5,11 @@ $BINPATH/euRun -n Ex0RunControl &
...
-$BINPATH/euCliMonitor -n Ex0Monitor -t my_mon & 
+$BINPATH/euCliMonitor -n Ex0Monitor -t my_mon &
 $BINPATH/euCliCollector -n Ex0TgDataCollector -t my_dc &
 # The following data collectors are provided if you build user/eudet
-#$BINPATH/euCliCollector -n DirectSaveDataCollector -t my_dc &
-#$BINPATH/euCliCollector -n EventIDSyncDataCollector -t my_dc &
-#$BINPATH/euCliCollector -n TriggerIDSyncDataCollector -t my_dc &
+$BINPATH/euCliCollector -n DirectSaveDataCollector -t my_dc &
+$BINPATH/euCliCollector -n EventIDSyncDataCollector -t my_dc &
+$BINPATH/euCliCollector -n TriggerIDSyncDataCollector -t my_dc &
```
lhuth, line 13: The -t <> option has to provide a unique name.

user/example/misc/scan/ExampleScan2.scan
lhuth: got it

gui/include/scanHelper.hh
```diff
@@ -5,6 +5,7 @@
 #include <string>
 #include <vector>
 #include <cstdlib>
+#include <map>
```
lhuth, line 8: at least on my machine `#include <unordered_map>` is required to compile

lhuth (changes requested):

user/example/misc/scan/ExampleScan2.scan
```diff
@@ -0,0 +1,40 @@
+[global]
```
lhuth, line 1: Maybe you can give the multi parameter scan file a more meaningful name?

user/example/misc/example_startrun.sh
kraemeru: I don't remember doing this change but maybe I simply misremember? But I now called them something else specific  to the type of data collector.

lhuth (approved):

PR #690, opened 2022-11-10
Caribou apts calib

ffeindt:
Recent developments on the AD9249converter adding the possibility to define calibration functions in the corryvreckan configuration file.


user/caribou/module/src/AD9249EventConverter.cc
```diff
@@ -65,20 +74,44 @@ void AD9249Event2StdEventConverter::decodeChannel(
...
     m_useTime = conf->Get("use_time_stamp", false);
     m_waveform_filename = conf->Get("waveform_filename", "");
 
-    EUDAQ_DEBUG( "Using configuration:" );
-    EUDAQ_DEBUG( " threshold_low  = " + to_string( threshold_low ));
-    EUDAQ_DEBUG( " threshold_trig  = " + to_string( threshold_trig ));
-    EUDAQ_DEBUG( " use_time_stamp  = " + to_string( m_useTime ));
+    // read calibration functions
+    delete m_calib_functions.at(
+        0); // since there is only one 'name' this destroys them all
```
simonspa, line 94: Oh wow, that is quite a nice example of very sketchy memory management :/

Is there a reason you really need them to be pointers? Otherwise

```cpp
std::vector<TF1>
AD9249Event2StdEventConverter::m_calib_functions{};

m_calib_functions.clear();
for (unsigned int i = 0; i < m_calib_strings.size(); i++) {
    m_calib_functions.emplace_back(name.c_str(), m_calib_strings.at(i).c_str(), m_calib_range_min, _calib_range_max);
}
```

would be much cleaner and also more safe (right now you loop over `m_calib_strings` and use their index to access elements in `m_calib_functions` - this will fail (with `std::out_of_bounds`) if `m_calib_strings.size() != m_calib_functions.size()`.

simonspa (changes requested):

user/caribou/module/src/AD9249EventConverter.cc
ffeindt: Yup, this is the better solution!

user/caribou/module/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
 aux_source_directory(src MODULE_SRC)
 INCLUDE_DIRECTORIES(include)
 
+find_package(ROOT REQUIRED)
```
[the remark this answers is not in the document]
simonspa: Implemented solution still to be tested...

user/caribou/module/CMakeLists.txt
```diff
@@ -1,9 +1,20 @@
 aux_source_directory(src MODULE_SRC)
 INCLUDE_DIRECTORIES(include)
 
+find_package(ROOT)
+
+if(NOT ROOT_FOUND)
```
simonspa, line 6: To be tested.

user/caribou/module/CMakeLists.txt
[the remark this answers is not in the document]
ffeindt: well this does not build, if I do not source root. is this what you had in mind as test? it also fails for master though. or does it require some cmake flag to build without root? The error I am getting is:
```
CMake Error at main/lib/core/CMakeLists.txt:11 (ROOT_GENERATE_DICTIONARY):
  Unknown CMake command "ROOT_GENERATE_DICTIONARY".
```

user/caribou/module/src/AD9249EventConverter.cc
```diff
@@ -65,20 +75,42 @@ void AD9249Event2StdEventConverter::decodeChannel(
...
+      m_calib_functions.emplace_back(name.c_str(), m_calib_strings.at(i).c_str(),
+                                     m_calib_range_min, m_calib_range_max);
+    }
+
+    EUDAQ_DEBUG("Using configuration:");
+    EUDAQ_DEBUG(" threshold_low  = " + to_string(threshold_low));
+    EUDAQ_DEBUG(" threshold_trig  = " + to_string(threshold_trig));
+    EUDAQ_DEBUG(" use_time_stamp  = " + to_string(m_useTime));
+    EUDAQ_DEBUG("Calibration functions: ");
+    for (unsigned int i = 0; i < m_calib_strings.size(); i++) {
```
simonspa, line 108: Does EUDAQ tell you somehow in which debug level it is? Then you woudn't have to do 16 loop iterations when nothing if printed (Corry has e.g. `IFLOG(DEBUG) { /*...*/ }`)

user/caribou/module/src/AD9249EventConverter.cc
```diff
@@ -224,12 +265,12 @@ bool AD9249Event2StdEventConverter::Converting(
   d2->AddPlane(plane);
 
   // use timestamps
-  if( m_useTime ){
+  if (m_useTime) {
```
simonspa, line 268: Is this still necessary of superseded by the new upstream functionality https://gitlab.cern.ch/corryvreckan/corryvreckan/-/blob/master/src/modules/EventLoaderEUDAQ2/README.md?plain=1#L99 ?

user/caribou/module/CMakeLists.txt
ffeindt: Fixed and tested with and without root.

user/caribou/module/src/AD9249EventConverter.cc
```diff
@@ -21,10 +21,20 @@ int AD9249Event2StdEventConverter::threshold_trig(1000);
...
+// baseline evaluation
+int AD9249Event2StdEventConverter::m_blStart(200);
+int AD9249Event2StdEventConverter::m_blEnd(100);
+// calibration functions
+double AD9249Event2StdEventConverter::m_calib_range_min(0);
+double AD9249Event2StdEventConverter::m_calib_range_max(16384);
+std::vector<std::string> AD9249Event2StdEventConverter::m_calib_strings(16,
+                                                                        "x");
+std::vector<TF1>
+AD9249Event2StdEventConverter::m_calib_functions(16, TF1("name", "x"));
```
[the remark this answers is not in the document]
ffeindt: Done for the TF1, the string is used as default value, if no function is given in the configuration file.

user/caribou/module/src/AD9249EventConverter.cc
ffeindt: It does `EUDAQ_IS_LOGGED`.

user/caribou/module/src/AD9249EventConverter.cc
ffeindt: Yes it is. It will change the default behavior, but since this affects only us I guess it is fine.

user/caribou/module/include/CaribouEvent2StdEventConverter.hh
```diff
@@ -2,6 +2,9 @@
 #include "eudaq/RawEvent.hh"
 #include "eudaq/Logger.hh"
 
+// Foreward declaration of TF1, so that the header has no root dependecies
```
simonspa, line 5: Nice :smirk_cat:

user/caribou/module/include/CaribouEvent2StdEventConverter.hh
```diff
@@ -36,12 +40,26 @@ namespace eudaq {
     void decodeChannel(const size_t adc, const std::vector<uint8_t>& data, size_t size, size_t offset, std::vector<std::vector<uint16_t>>& waveforms, uint64_t& timestamp) const;
     static size_t trig_;
```
simonspa, line 41: The `trig_` could be removed with a) reading it from the FPGA event ID and b) the new filter in Corry upstream?

user/caribou/module/src/AD9249EventConverter.cc
```diff
@@ -4,55 +4,74 @@
...
+      // The timestamp spans multiple samplings and the sampling where it starts
+      // is marked with 0x1 in the status bits. The status bits are carried in
+      // Channel 7, i.e. the last one - which means only after reading and
+      // storing 7x2 bits of the timestamp already, we know if they were at the
+      // begin of a new timestamp. If not, we need to rewind the index and start
+      // over again unti we find a timestamp start.
       if (ts_i < 8 && (ts & 0x1) == 0) {
         ts_i = 0;
+        ts_lost++;
+        timestamp = 0;
```
simonspa, line 74: Excellent finding.

simonspa (approved):

user/caribou/module/include/CaribouEvent2StdEventConverter.hh
ffeindt: a) I do not think we have that for this device. One could use the trigger ID from the EUDAQ event, but this would not work for our case since we store data from 2 devices in one file.
b) The other way around. This works only with the new filter in corry, since we do not increment it from the converter for the other device.
c) Since time stamps are fine now, all of that does not really matter anymore.

simonspa: d) excellent. Thanks!

PR #691, opened 2022-11-16
Unbreak EUDAQ2: Allow Compilation without ROOT

simonspa:
PR #656 added a direct ROOT dependency to the EUDAQ2 core library without even checking previously if ROOT is installed and has been found by the build system. Seems like we were all lucky and had ROOT installed anyway.

...unless we tried to build for an embedded 32-bit ARM system without ROOT :wink: 

I now quietly check for ROOT presence and enable dictionary generation in case we have it, and not else. This should be fine, because for using the resulting ROOT dicts you need ROOT anyway. Sorry, circular sentence.

(cc) @ffeindt @herkerta


lhuth:
or nobody updated EUDAQ 📦

lhuth (approved):

PR #696, opened 2022-12-16
Remove MuPix8 module

stephanlachnit:
This is required for newer versions of the MuPix8_DAQ.

PR #697, opened 2022-12-16
Adjust Producer.hh to support MuPix8_DAQ

stephanlachnit:
Unfortunately compiling MuPix8 with EUDAQ support is only possible with this patch. Separate to #696 since this changes core components and I'm not sure what the policy here is.


lhuth:
Hey, 
Good question -  I personally do not care too much about this :shrug: Maybe others have strong feelings. If not we can add this.

lhuth:
3 months of silence are interpreted as no hard feelings by anyone. Let's merge it.

lhuth (approved):

PR #699, opened 2023-01-16
dsipm updates

ffeindt:
* Added configuration option to check valid bit and time stamp consistency. If inconsistent event is dropped. Defaults to off.
* Added configuration option to set an effective number of bits for fine time-stamp decoding. Defaults to 32, the nominal number of bits.
* Always forwarding event time stamps. Event building by trigger ID can be forced in corry.
* Removing commented code and cleaning up debugging output.

### DIFF
--- a/user/caribou/module/include/CaribouEvent2StdEventConverter.hh
+++ b/user/caribou/module/include/CaribouEvent2StdEventConverter.hh
@@ -62,6 +62,7 @@ namespace eudaq {
   private:
     static bool m_configured;
     static bool m_zeroSupp;
+    static bool m_checkValid;
     static uint64_t m_trigger;
     static uint64_t m_frame;
   };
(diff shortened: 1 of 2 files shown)


user/caribou/module/include/CaribouEvent2StdEventConverter.hh
```diff
@@ -60,10 +60,13 @@ namespace eudaq {
     bool Converting(eudaq::EventSPC d1, eudaq::StandardEventSP d2, eudaq::ConfigurationSPC conf) const override;
     static const uint32_t m_id_factory = eudaq::cstr2hash("CariboudSiPMEvent");
   private:
+    uint8_t getQuadrant(const uint16_t& col, const uint16_t& row) const;
```
simonspa, line 63: Could even be

```cpp
static uint8_t getQuadrant(const uint16_T& col, const uint16_t& row);
```

since it does not rely on any member variables of `dSiPMEvent2StdEventConverter`. More a style question though

simonspa (dismissed):
I haven't checked all the calculations of the timestamp, but looks good to me overall.

user/caribou/module/include/CaribouEvent2StdEventConverter.hh
ffeindt: > I haven't checked all the calculations of the timestamp, but looks good to me overall.
Inge checked them and asked me to implement the possibility to make the effective number of bits for the fine TDC configurable and quadrant dependent, which I did here. Otherwise she had no complains. I also tested this version and results are the same as before the change, when one uses the default number of effective bits.

> Could even be `static uint8_t getQuadrant(const uint16_T& col, const uint16_t& row);` ...
Changed.

simonspa (approved):

PR #700, opened 2023-01-23
Allow to run multiple CaribouProducers with the same device type

simonspa:
The Caribou device name is also taken as the producer name (supplied via the `-t` CLI argument). If multiple producer operating the same device type in different Caribou setups are supposed to be operated, an identifier can be appended to the device name, separated by an underscore, e.g. `CLICpix2_upstream`. The CaribouProducer will strip everything after the last underscore from the device name before attempting to instantiate the device object.

@ffeindt @gianpierovignola @stephanlachnit required for our next testbeam with 2x dSiPM, please test.


stephanlachnit:
Tested with @gianpierovignola and works as expected

stephanlachnit (approved):

lhuth (approved):

Issue #709, opened 2023-04-05
macOS compilation issue: libcore_rdict.pcm missing


msuljic:
Error on `make install` on macOS 13:
```
Install the project...
-- Install configuration: "RelWithDebInfo"
-- Installing: /Users/msuljic/projects/eudaq/include/eudaq/Config.hh
-- Up-to-date: /Users/msuljic/projects/eudaq/cmake/eudaqTargets.cmake
-- Up-to-date: /Users/msuljic/projects/eudaq/cmake/eudaqTargets-relwithdebinfo.cmake
-- Up-to-date: /Users/msuljic/projects/eudaq/cmake/eudaqConfig.cmake
-- Up-to-date: /Users/msuljic/projects/eudaq/cmake/eudaqConfigVersion.cmake
CMake Error at main/lib/core/cmake_install.cmake:41 (file):
  file INSTALL cannot find
  "/Users/msuljic/projects/eudaq/build/main/lib/core/libcore_rdict.pcm": No
  such file or directory.
Call Stack (most recent call first):
  main/lib/cmake_install.cmake:42 (include)
  main/cmake_install.cmake:42 (include)
  cmake_install.cmake:76 (include)
```

This is the contents of `/Users/msuljic/projects/eudaq/build/main/lib/core`:
```
CMakeFiles              Makefile                cmake_install.cmake     eudaqCoreDict.cxx       libeudaq_core.2.5.dylib libeudaq_core.dylib     libeudaq_core.rootmap   libeudaq_core_rdict.pcm module.modulemap
```

I think the problem is that filename is hardcoded in `main/lib/core/CMakeLists.txt` L41, introduced in #691.

Unfortunately, I'm not proficient in cmake to fix it.

PR #710, opened 2023-04-06
Fix cmake issue of sharedlibrary loadpath in MacOS

msuljic:
This should fix https://github.com/eudaq/eudaq/issues/709 (unsure if it is the most elegant way).


simonspa:
Hi @msuljic 

I was just checking in other projects we have that cross-compile also on macOS - and we never set these values at all but leave them to the CMAKE-internal defaults. Do things work if you just remove the two lines entirely?

Cheers,
Simon

msuljic:
Yes they do - update pushed!

PR #711, opened 2023-04-09
user ADENIUM folder structure and cleanup

arummler:
Changed the ADENIUM folder structure to current conventions (nested; converter plugin in module folder). I don't think it will need a dependency on the CMS software.


simonspa:
Thanks for noticing and removing the wrong linking directories.

However, for the rest:

- i think the current "convention" is unnecessary and overly complex. While I wouldn't change existing producers, i don't see the need to stick with them.
- the name was chosen intentionally. There is a separate module containing the actual producer code, and after your rename they would clash and overwrite each other.

lhuth:
Hi, 
I do not have a strong opinion on the naming scheme, but the second point of Simon is essential.

@arummler Mind to update? If not we can also close this one

lhuth:
No news on this - @arummler?

lhuth:
>12 months of silence... going to close.

arummler:
I will reopen at a later stage.

PR #713, opened 2023-04-16
EUDAQ version accessible through API

arummler:
Introducing the same API as I did in eudaq v1 for retrieving the eudaq version information inside a consuming project. E.g. in the YARR producer I am taking this information and writing it into the BORE for documentation. Mechanism is based on "Professional CMake" chapter 21.


simonspa:
Hi @arummler 

in my opinion a bit elaborate, but if you need this downstream, alright.

As it is now it breaks compilation (or rather: configuration) on systems without `git` executable installed. I would suggest to make all of the fancy "tweak", "branch" et cetera stuff optional only if git has been found. Otherwise they can still return an empty string or zero as version numbers.

arummler:
Right, that case is not covered. I will fix it. Also I might make use of the lib versioning done in the main CMakeLists.txt which I just spotted; let me check it.

Issue #712, opened 2023-04-12
#710 broke StdEventMonitor on Linux/CentOS


msuljic:
#710 while appeared to resolve compilation issues on MacOS, broke StdEventMonitor on CentOS.

Command
```
$ StdEventMonitor -d run497194033_221211201609.raw -l DEBUG -e 100 -E 1000
```
yields error messages like this for every event:
```
[2023-04-12 15:51:01.630] (WARN) [unknown sender]WARNING, no StdEventConverter for RawEvent with ExtendWord(1605811347) and Description(ITS3global)
Factory<0x7fe4dd2df000>:  Unknown class ID: <1605811347>
```
Reverting to latest commit before the PR solves the issue on CentOS (ofc, compilation on MacOS will likely revert to be broken).

Additional info - build commands:
```
$ sh make_clean.sh && cd build
$ cmake -DEUDAQ_BUILD_STDEVENT_MONITOR=ON -DUSER_ITS3_BUILD=ON -DEUDAQ_BUILD_PYTHON=ON ..
$ make -j 10 && make install
```

simonspa:
Hi @msuljic 

thanks for reporting this - and sorry @ffeindt that this caused you to dig around why your converters don't work anymore.

I investigated what the reason is and how we could solve it. Essentially the eudaq build system here *abuses* a variable that CMake uses to add prefixes (like `lib`) to adjust library names to the host system. Now EUDAQ just overwrites it to add its `eudaq_` prefix...

My approach that would not involve reworking large fractions of the build system to properly add the prefix to the target names instead of to the library prefix would be this: https://github.com/eudaq/eudaq/pull/714

Cheers,
Simon

stephanlachnit:
Yeah I can confirm #710 breaks everything... Can we revert until we have a proper fix?

simonspa:
@stephanlachnit care to fix? :)

PR #714, opened 2023-04-20
CMake: Only Append eudaq_ to Library Prefix

simonspa:
This appends `eudaq_` to an existing (platform-dependent) library prefix instead of overwriting it.

Needs testing on macOS, @msuljic any chance you could look at his?


msuljic:
Hi @simonspa,

unfortunately I cannot test it at the moment on macOS as after updating to the latest Xcode (14.3) I cannot compile eudaq at all any more (some header files changed and other projects also experience similar problems, but there is not solution yet).

While I'm trying to fix it, I can tell you that I don't believe that this will work due to this line: https://github.com/eudaq/eudaq/blob/07d976b063b1fae53c89425e8ceae45da932ea22/main/lib/core/CMakeLists.txt#L41
The name `libcore_rdict.pcm` is hardcoded, while due to the `CMAKE_SHARED_LIBRARY_PREFIX` `eudaq_core_rdict.pcm` will be created (same problem as in #709.

lhuth:
Compiles on mac os and unix 🚀

lhuth (approved):

msuljic:
> Compiles on mac os and unix 🚀

Compiles and works or only compiles? I.e. checking that we don't run into #712 again...

simonspa:
Dear @msuljic,

thank you very much for your comment. Testing the changes on your end would have been about the same amount of work and we would really appreciate this minimal effort if we are already fixing your issues.

msuljic:
> Dear @msuljic,
> 
> thank you very much for your comment. Testing the changes on your end would have been about the same amount of work and we would really appreciate this minimal effort if we are already fixing your issues.

Hi, thank you for fixing MY issues. I am very glad to help and do the testing, however, if it was already tested, testing on my end would have doubled the amount of work. :) I unfortunately cannot test on MacOS any more. On CentOS stream 8 it doesn't seem to work, at least not out of the box (same issue as #712).

lhuth:
Hi, 

tested on unix (ubuntu 22.04.1 & centOS 7) and on macOS ventura 13.4.1. Monitor fully functional on both. The issue reported in #712 includes
```
[2023-04-12 15:51:01.630] (WARN) [unknown sender]WARNING, no StdEventConverter for RawEvent with ExtendWord(1605811347) and Description(ITS3global)
Factory<0x7fe4dd2df000>:  Unknown class ID: <1605811347>

```
This indicates that the ITS3global has not been built correctly as you cannot find the converter for `Description(ITS3global)`


Can you start off with a completely fresh build (manually delete `build` `lib` and `bin`)

msuljic:
Apologies, it works, but due to my haste I misread the error message while the result i.e. the monitor screen looked identical.

A bit longer answer:
Here in main repo this line https://github.com/eudaq/eudaq/blob/d25579a35ead4f89a925714d1c5ca44d007a4e65/monitors/onlinemon/src/OnlineMon.cxx#L180 makes sure I don't get any event decoded as due to many planes in my setup, first 10 events are status events containing 0 planes. Which results in error:
```
Event #9668 has 6 plane(s), while we expect 0 plane(s).  (Event is skipped)
Factory<0x7f7920db6000>:  Unknown class ID: <3866605352>
```
Repeated indefinitely. Similar, but sufficiently different that I should have spotted it. Sorry about that.

PR #715, opened 2023-04-26
Caribou/dSiPM: latest testbeam updates

stephanlachnit:
@simonspa @ffeindt

### DIFF
--- a/user/caribou/README.md
+++ b/user/caribou/README.md
@@ -84,3 +84,7 @@ The following parameters can be passed in the configuration in order to influenc
 The following parameters can be passed in the configuration in order to influence the decoding behavior of this module:
 
 * `plane_id`: Value that will be set `eudaq::StandardPlane::ID`. This setting can be used to support multiple devices in the same run. Default is `0`.
+* `discard_during_reset`: If `0`, frames that appear during the frame reset will not be ignored. Default is `1`.
+* `check_valid`:
+* `frame_start`: First frame to be used in an event for a given trigger. Can be set to a higher value to ignore the first frame(s). Defaults to `0`.
+* `frame_stop`: Last frame to be used in an event for a given trigger. Can be set to a lower value to ignore the last frame(s). Defaults to `2`. Might need to be adjusted depending on the peary configuration.
--- a/user/caribou/module/src/dSiPMEventConverter.cc
+++ b/user/caribou/module/src/dSiPMEventConverter.cc
@@ -11,49 +11,71 @@ namespace {
       dSiPMEvent2StdEventConverter>(dSiPMEvent2StdEventConverter::m_id_factory);
 }
 
-bool dSiPMEvent2StdEventConverter::m_configured(0);
-bool dSiPMEvent2StdEventConverter::m_zeroSupp(1);
-bool dSiPMEvent2StdEventConverter::m_checkValid(0);
-uint64_t dSiPMEvent2StdEventConverter::m_trigger(0);
-uint64_t dSiPMEvent2StdEventConverter::m_frame(0);
-double dSiPMEvent2StdEventConverter::m_fine_ts_effective_bits[4];
+std::vector<bool> dSiPMEvent2StdEventConverter::m_configured({});
+std::vector<bool> dSiPMEvent2StdEventConverter::m_zeroSupp({});
+std::vector<bool> dSiPMEvent2StdEventConverter::m_discardDuringReset({});
+std::vector<bool> dSiPMEvent2StdEventConverter::m_checkValid({});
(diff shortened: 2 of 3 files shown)


user/caribou/README.md
```diff
@@ -84,3 +84,7 @@ The following parameters can be passed in the configuration in order to influenc
 The following parameters can be passed in the configuration in order to influence the decoding behavior of this module:
 
 * `plane_id`: Value that will be set `eudaq::StandardPlane::ID`. This setting can be used to support multiple devices in the same run. Default is `0`.
+* `discard_during_reset`: If `0`, frames that appear during the frame reset will not be ignored. Default is `1`.
+* `check_valid`:
```
simonspa, line 88: Undocumented? :smile:

user/caribou/module/src/dSiPMEventConverter.cc
```diff
@@ -11,50 +11,72 @@ namespace {
       dSiPMEvent2StdEventConverter>(dSiPMEvent2StdEventConverter::m_id_factory);
 }
 
-bool dSiPMEvent2StdEventConverter::m_configured(0);
-bool dSiPMEvent2StdEventConverter::m_zeroSupp(1);
-bool dSiPMEvent2StdEventConverter::m_checkValid(0);
-uint64_t dSiPMEvent2StdEventConverter::m_trigger(0);
-uint64_t dSiPMEvent2StdEventConverter::m_frame(0);
-double dSiPMEvent2StdEventConverter::m_fine_ts_effective_bits[4];
+std::vector<bool> dSiPMEvent2StdEventConverter::m_configured({});
```
simonspa, line 14: Probably it would be cleaner to bake them all into a `struct` or `class`.

user/caribou/module/src/dSiPMEventConverter.cc
```diff
@@ -126,29 +152,32 @@ bool dSiPMEvent2StdEventConverter::Converting(
...
-      EUDAQ_WARN("Fine clock == 31. Interesting!");
+      EUDAQ_WARN("Coarse clock == 0. This might screw up timing analysis.");
+      if (m_discardDuringReset[plane_id] == true) {
+        return false;
+      }
     }
+    // According to Inge this needs to be discarded
     if (clockFine == 0) {
-      clockFine = 32;
+      EUDAQ_WARN("clockFine == 0. This might screw up timing analysis.");
```
simonspa, line 177: shouldn't this warning only be printed if we do *not* discard the frame?

user/caribou/module/src/dSiPMEventConverter.cc
```diff
@@ -210,7 +238,10 @@ bool dSiPMEvent2StdEventConverter::Converting(
   // Store frame begin and end in picoseconds
   d2->SetTimeBegin(frameStart);
   d2->SetTimeEnd(frameEnd);
-  d2->SetTriggerN(trigger_id_fpga);
+  d2->SetTriggerN(fpgadata.trigger_id);
+
+  // Add tag with frame number for additional information
+  d2->SetTag("frame", m_frame[plane_id]);
```
simonspa, line 244: In principle, the `StandardPlane` class object generated in L126 already has the feature of storing multiple frames (implemented for M26), see https://github.com/eudaq/eudaq/blob/master/main/lib/core/include/eudaq/StandardPlane.hh#L117

I'm not sure how useful this is for us, but maybe have a look to at least check if it makes sense.

user/caribou/module/src/dSiPMEventConverter.cc
stephanlachnit: Yup, looks significantly better now.

user/caribou/module/src/dSiPMEventConverter.cc
stephanlachnit: Hm, apparently we can attach this to each pixel: https://github.com/eudaq/eudaq/blob/2b22f1472498ffd0cc8d16a4afea52b26c24342f/main/lib/core/include/eudaq/StandardPlane.hh#L54

I have to look a bit where we can get this back. The idea was that we might be able to use this in corry to filter for a tag, but in the end, we just added it directly in the convert via `frame_start` and `frame_stop`.

stephanlachnit: Ok looking at this:
https://github.com/eudaq/eudaq/blob/2b22f1472498ffd0cc8d16a4afea52b26c24342f/main/lib/core/src/StandardPlane.cc#L288

I don't seem to understand what this actually does...

simonspa: It's fine, I also like the tag - useful info anyway. :)

simonspa (approved):

PR #717, opened 2023-07-26
Update cmake FindPython

cbespin:
`FindPythonLibs` is [deprecated](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) since `cmake==3.12`. Since Ubuntu 18 that was shipped with `cmake < 3.12` is EOL since April 2023, it might make sense to update to the newer `FindPython`.

Tested and works on Debian 12, cmake 3.25, python 3.10.


stephanlachnit (approved):
LGTM /cc @lhuth

simonspa (approved):

PR #718, opened 2023-08-08
Caribou/dSiPM latest changes

stephanlachnit:
Latest changes for dSiPM Event Decoder:

#### Discard during Reset Updates

After some analysis the best filter method was evaluated to just check `bunchCounter==0` and `clockFine == 0`, removing the `clockCoarse==0` check since this doesn't seem to happen without one of the other two cases. The motivation is that while `bunchCounter==0` and `clockFine == 0` don't make sense in a testbeam measurement, `clockCoarse==0` can be a valid value.

#### Implemented DNL Correction

This replace `fine_ts_effective_bits_qX` with a more powerful `fine_tdc_bin_widths_qX`, which is an array defining the fine TDC bin width. The new timestamps are set by summing the fine TDC bins < the counter value + 0.5 * the fine TDC bin width that was measured (i.e. placing it in the middle of the bin).

Idk if it's worth to keep a backup for `fine_ts_effective_bits_qX`, since it is a bit simpler than setting an array with the same fine TDC bin widths. To be decided by @ffeindt.

The parser for the arrays is not the best in the world (you can break it if you want), unfortunately there is no built-in way to parse arrays in eudaq configs. I think it's good enough though since we don't server a large userbase anyway.

#### Implemented Pixel Delays

Pixel Delays are subtracted from the timestamp if given a pixel delay map in the configuration e.g. from a laser measurement.

Implemented with the same parser described above, however now with 32x32 values. Not particularly intuitive so this is something where you would write a script to output the config string.

#### Misc

- Bumped the default value of `frame_stop` to 127 so it doesn't need to be set explicitly if you want to look at all (three or more) frames.
- Added missing counter rollover for `clockCoarse == 0` and added it back again for `clockFine == 0`

/cc @ffeindt @simonspa


simonspa:
Looking good to me. Any blockers to merge?

ffeindt:
Not from my side!

PR #731, opened 2023-10-28
add simple alibava producer which calls external alibava-gui

dsperlich:
and optionally extern HV setting programm
 
 We currently run this with an alibava testbeam at DESY


simonspa (approved):
LGTM

PR #735, opened 2023-11-19
port gui to Qt6 (keeping Qt5 as fallback)

arummler:
- updated a but the Qt5 cmake
- will use by default Qt6 but will fall back to Qt 5.15 (last LTS)


lhuth:
As Simon said - one should not merge his own changes without review. This causes a potential  unstable version. Please do not merge your own stuf!

Issue #703, opened 2023-03-10
EUDAQ2: dynamic exception error (error: ISO C++17 does not allow dynamic exception specifications) 


ajitkmaurya:
Hello,

I am trying to compile eudaq2 with `-DEUDAQ_LIBRARY_BUILD_LCIO=ON`. But it gives me following error.

`/eudaq2/extern/include/EVENT/LCEvent.h:66:74: error: ISO C++17 does not allow dynamic exception specifications`
`66 |     virtual LCCollection * getCollection(const std::string & name) const throw (DataNotAvailableException, std::exception )  = 0;`

Could you please suggest how to solve this.

EUDAQ2 version: 2.5.2
OS: Ubuntu 22.04

Thanks 
Ajit

simonspa:
The issue is that (for some reason) EUDAQ bundles a very outdated zipped version of the LCIO library here: https://github.com/eudaq/eudaq/tree/master/main/lib/lcio/extern
This should definitely be removed @lhuth @herkerta 

But in the meantime, you can do the following:

* get a recent copy of LCIO (from here: )
* Compile it with
   ```
   mkdir build && cd build/
   cmake -DCMAKE_INSTALL_PREFIX=/path/to/your/eudaq/extern/ ..
   make install -j
   ```
* Then recompile EUDAQ with clean build directory and it should find your externally built library.

ajitkmaurya:
Thank you! 
It works.

simonspa:
Please don't close, this is just a workaround

PR #740, opened 2024-02-04
Update LCIO library

arummler:
- remove integrated old source file version
- get new v02-20-02 through fetch_content 
- use proper targets (after upstream lcio fix)
- not sure whether the RPATH patch is still needed; if yes there is now an option LCIO_SET_RPATH that can be simply set
- should fix https://github.com/eudaq/eudaq/issues/703


simonspa:
Looking good, thanks

simonspa (approved):

PR #741, opened 2024-02-05
Revert "port gui to Qt6 (keeping Qt5 as fallback)"

arummler:
Reverts eudaq/eudaq#735
Okay, I was a bit hasty. Offering to revert and will then re-do the PR some other time in the future.


simonspa:
I reset the branch to before your merge. Please refrain from merging code into `master` without prior review from a second party.

PR #738, opened 2024-02-02
CI: cross platform re-implementation as github actions

arummler:
Summary of changes:
- CI implemented for linux gcc and clang, macos clang and windows msvc using github actions
- Improved C++ version detection using ROOT C++ version
- Fixed FileSystem linking
- fixed test on Windows
- removal of appveyor

PR #742, opened 2024-02-06
port gui to Qt6 (keeping Qt5 as fallback)

arummler:
- updated a bit the Qt5 cmake
- will use by default Qt6 but will fall back to Qt 5.15 (last LTS)
- possible to force Qt5
- included in CI / ubuntu runner


arummler:
To be rebased and merged after #738

herkerta (approved):

PR #743, opened 2024-02-14
YARR converter plugin  and MMC3 (stcontrol gen3) converter plugin

arummler:
* added converter pluging for MMC3 capable version of STcontrol which I ported to eudaq2 (to be merged with qt6 development)
   * https://gitlab.cern.ch/arummler/USBpix/-/tree/master_development_may_2022?ref_type=heads
* added YARR (ATLAS ITk pixel DAQ) converter plugin
   * https://gitlab.cern.ch/YARR/YarrEudaqProducer/-/tree/eudaq2-api-update?ref_type=heads
* extended the StdEvtMonitor to accept both
* has been running at CERN H6 for the last two years like that


arummler:
To be merged after #738 (and rebase as needed).

arummler:
@dinardo I had one small conflict here, the rest was compatible as I believe part of it came from the run control pc in H6. You REF has 16 or 32 LVL1 bins? I assumed that this is also a CROC and hence 32. Please check line 85 in HitmapHistos.cc

lhuth (approved):
This is user code that I will not interfere with. And according to @arummler in use for a year

PR #748, opened 2024-04-27
Fixed a bug in the eudaq scan that did not allow for negative scan st…

kraemeru:
…epping. Also added a debug output that tells you some more information about start and end point as well a step differences when running in parallel.
 
Also fixed a bug that in a negative scan step the original code already converted it to negative when stop < start which then resulted in it running in the wrong direction. Now it converts step to absolute meaning the user can enter either negative or positive, as long as start and stop are correct it will always use the correct sign.


lhuth (approved):
merci - lgtm

PR #749, opened 2024-05-08
Allow command receivers to load INI/CONF files

forthommel:
This PR introduces two new `LoadInitConfiguration` and `LoadConfiguration` protected members to the `CommandReceiver` base object, allowing for its derivative to... load INI and CONF configuration fragments and steer the module directly. The command receiver loop was slightly modified to use these two methods and avoid any code duplication.

This allows the ROOT monitoring tool to run in standalone mode and still be configured according to user's needs when reprocessing runs collected in the past, thanks to a new UI button, and a readout of the user-steered file.

This PR is not expected to be as large as it is, but all affected files were filtered through a `clang-format` run.


forthommel:
Friendly reminder!

simonspa:
Hi @forthommel - to whom?

This is a fairly large change to a framework nobody really actively develops on, so this might just take some time until someone steps up and review this.

/Simon

forthommel:
Hello @simonspa!
Just reaching out to know whether you found some additional manpower to work on the core of eudaq2, or if this can be considered as stalled?

simonspa:
Hi @forthommel 

currently there is not more personpower for EUDAQ2, especially since we are working on a "successor" framework which is newly written from ground up but should provide a similarly easy integration interface for users. Feel free to contact me for more info.

/SImon

PR #750, opened 2024-05-14
H2M: add frame ID back-jump detection logic

simonspa:
Still to be tested by CERN.


ffeindt (commented):
I suppose this works for CERN? Otherwise there would have probably been further discussions?

PR #752, opened 2024-06-19
Fixing root C++ version detection for new root installs. 

arummler:
Inspired by commit a58d850 in allpixsquared.


simonspa:
linking doesnt hurt :) then there is more info

simonspa (approved):

PR #753, opened 2024-07-19
Update pybind version from 2.5.0 to 2.13.1

msuljic:
This suppresses a number of pybind related warnings during compilation.


arummler (commented):
Should be okay. Although we might want to renovate that cmake file at some point (and also build wheels as discussed once). Also: shouldn't we remove the old pybind versions?

msuljic:
@arummler For cmake file unfortunately I cannot help. For old version I have no strong opinion. They stay in the git tree anyhow so I see no obvious (dis)advantage of doing it either way.

arummler (approved):
Sure. I had already played around with the CI/wheel building a couple years ago. Might come back to it if I have a chance. The old versions we can keep for the moment, will probably remove later.

PR #755, opened 2024-08-01
Fixing use of Caribou Peary configuration class in CaribouProducer

yotarid:
This PR relates to the Peary PR https://gitlab.cern.ch/Caribou/peary/-/merge_requests/205 where modification were made to the configuration class. It adapts for the modification in the CaribouProducer

### DIFF
--- a/user/caribou/CMakeLists.txt
+++ b/user/caribou/CMakeLists.txt
@@ -13,6 +13,6 @@ endif()
 
 include_directories(${EUDAQ_INCLUDE_DIRS})
 
-FIND_PACKAGE(Peary REQUIRED)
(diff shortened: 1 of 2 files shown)


yotarid:
The modifications in EUDAQ are not yet tested but should in principle be transparent. @simonspa Could you please confirm ? Otherwise I can test it sometime next week

simonspa:
Excellent, thanks, that slipped when changing the interface. I fixed one thing (using wrong config for secondary device), updated include paths and added a minimum version requirement.

In peta-caribou we should move to using a tagged version of peary.

simonspa (approved):

PR #757, opened 2024-09-30
Fix building PI stage controller

simonspa:
`PI_FOUND` contains the library path. Use it :)


simonspa:
PS: no idea how this ever worked, @lhuth

arummler (approved):
Should now work.

PR #759, opened 2024-10-01
forward declaration of TFile

lhuth:
Don't want to include root in Converter header

### DIFF
--- a/user/caribou/module/include/CaribouEvent2StdEventConverter.hh
+++ b/user/caribou/module/include/CaribouEvent2StdEventConverter.hh
@@ -5,12 +5,12 @@
 #include "utils/datatypes.hpp"
 #include "utils/log.hpp"
 #include "utils/utils.hpp"
-#include <TFile.h>
 #include <array>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


user/caribou/module/include/CaribouEvent2StdEventConverter.hh
```diff
@@ -5,12 +5,12 @@
 #include "utils/datatypes.hpp"
 #include "utils/log.hpp"
 #include "utils/utils.hpp"
-#include <TFile.h>
 #include <array>
 #include <vector>
 
-// Foreward declaration of TF1, so that the header has no root dependecies
+// Foreward declaration of TF1/TFile, so that the header has no root dependecies
```
arummler, line 11: Foreward --> Forward ;)
dependecies --> dependencies

simonspa, line 11:
```suggestion
// Forward declaration of TF1/TFile, so that the header has no root dependencies
```

simonspa (approved):

PR #761, opened 2024-10-16
Online Monitoring tool utilising corryvreckan

s6anloes:
Addition of a new tool which allows monitoring of test beam data recorded using EUDAQ2 as DAQ control system to be monitored (semi-)online by starting a corryvreckan session and utilising its OnlineMonitor module. Since this project also required changes in corryvreckan, it only works with my recent changes to corry for which I am creating a merge request in parallel.

There is also a new file reader and deserialiser to read remote files connected via XRootD. Implemented as user module, because it requires the XRootD library and should therefore only be compiled when needed.

_Note:_ This is an update from a previous merge request that got outdated.


lhuth (commented):
Fine with me - but i am biased....

simonspa:
Are the breaking changes in the core lib necessary?

user/xrootd/module/CMakeLists.txt
arummler, line 1: I do not have time today to look through in detail. But regarding the cmake style: afaik xrootd 5.6... exports a proper cmake module which can be used together with target linking. As alternative a finder module (which certainly already someone wrote) which defines a target can be included but that is old fashioned. And should the case that ROOT has been found and that it provides an inbuilt xrootd be supported?

arummler:
> Are the breaking changes in the core lib necessary?

I would not call it breaking, more extension and bug fix.
I see:
- adding the SectionList
- more file error handling
- fixing getting the right config in the Monitor base class OnInit

For the latter: I do not think that anybody is relying on getting the - wrong - Configuration step config in the init step for a monitor. Neither ROOTmonitor nor StdEvtMonitor are using it at all. Quickly looking there are only the two non-production monitors accessing this info:
example/module/src/Ex0Monitor.cc
Dummy/module/src/DummyMonitor.cc
and they are actually implementing getting the right config file for the init step.

simonspa:
I indeed misread that part. Then fine with me.

monitors/corrymonitor/README_CorryMonitor.md
```diff
@@ -0,0 +1,103 @@
...
+**Please note that for the `corryvreckan` installation the `-DBUILD_EventLoaderEUDAQ2=ON` flag needs to be set to work with `CorryMonitor`.**
+
+## Installation
+
+To build the `CorryMonitor`, the `EUDAQ_BUILD_CORRYMONITOR` needs to be set to `ON` (default value is `OFF`).
+
+## Known Restrictions of the `CorryMonitor`
+
+- The CorryMonitor is currently exclusively running on Linux due to the use of *inotify*.
+- It is currently required to run the Monitor on the same machine as the RunControl to allow access to the `eudaq` configuration files. (The DataCollectors can run remotely via XRootD, explained below).
```
arummler, line 18: Is this correct? The config and init file sections should be read by the runcontroller and transmitted through the usual mechanism. Are you accessing them directly?

monitors/corrymonitor/src/CorryMonitor.cc
```diff
@@ -0,0 +1,592 @@
...
+  // Check if config for corryvreckan is found
+  struct stat buffer;   
+  if(stat(m_corry_config.c_str(), &buffer) != 0) {
+    std::string msg = "Config for corryvreckan cannot be found under "+m_corry_config+" ! Please check your /path/to/corryconfig.conf (Avoid using ~)";
+    EUDAQ_ERROR(msg);
+    EUDAQ_THROWX(eudaq::FileNotFoundException, msg);
+  }
+
+
+  // command to be exectued in DoStartRun(), stored tokenized in m_args.argv
```
arummler, line 220: executed

monitors/corrymonitor/src/CorryMonitor.cc
```diff
@@ -0,0 +1,592 @@
...
+        // Add each argument in turn, then the terminator.
+        while ((cstr = strtok (0, TOKEN)) != 0){
+          m_args.argv = addArg (m_args.argv, &m_args.sz, &m_args.used, cstr);
+        }
+      }
+    }
+
+    m_args.argv = addArg (m_args.argv, &m_args.sz, &m_args.used, 0);
+
+    /*
```
arummler, line 526: I think you want to remove this comment. Seems that was just a test?

monitors/corrymonitor/README_CorryMonitor.md
s6anloes: The problem is not the initialisation and configuration of the monitor, this gets passed as usual via the runcontrol.
The problem is the configuration for the datacollectors, because I need to get the file-naming pattern from there. But there is currently no mechanism to send the DataCollector-configuration to a Monitor module, I believe. So this where the limitation comes from. If it is not on the same machine, the CorryMonitor cannot access the information from the DataCollector configuration

user/xrootd/module/CMakeLists.txt
s6anloes: I agree that this would be the cleaner solution, but I had problems getting it to run. Locally I have a private XRootD installation, but for testing on lxplus I was using the key4stack to provide it.

I'll try to do some more testing, to see if I can use find_package()

s6anloes: I have updated the CMakeLists.txt to include the find_package method, and left the option to pass it manually.
From my testing I have found that using the Xrootd that root provides is not handled any different to the standalone one, so it should be covered with the update as well.

arummler: Okay for now. I believe there is also a target exported but cannot check now as the CI is not running until my PR is merged.

lhuth:
@arummler: Mind to check that one? would be cool to also have it in

arummler:
Sure, I will do. I just wanted to get the CI first passing again before new features get merged.

arummler:
I ran successfully the CI for Linux (xrootd module and corrymon) and Mac (xrootd). On Windows there is neither inotify nor afaik a xrootd build (although I saw a couple years ago an alpha version being discussed) On Mac there is no inotify. I will merge and then add CI modifications with another commit.
- We should mark it in CMAKE config for the non-caompatible operation systems.
- I think inotify can be replaced by other APIs on Mac and Windows. I found quickly a cross platform wrapper: https://github.com/emcrisostomo/fswatch but did not test it.
- We had this popping up already a couple times but I wonder whether in this case a eudaq monitor linked linked into a corry module (or with an intermediate binary) might not make it easier / avoid all this file monitoring issues. On the other hand it might not be able to do everything this solution can do.

PR #762, opened 2024-10-23
Fix return code of Adenium Decoder

simonspa:


simonspa:
Only relevant change that is not just debug messages is that Adenium should return `false` (failed decoding) when it doesn't find its required data blocks.

lhuth (approved):
Fine with me. Thanks for spotting. H2M part in there on purpose? Anyways fine for me. Feel free to merge.

simonspa:
Yes, on purpose. Also adds more debug logging there.

PR #764, opened 2024-12-04
H2M: ToT calibration

sararuizdaza:
Adding the possibility to read a ToT calibration file. 
If a ToT calibration file is not provided in the Corryvreckan configuration, the ToT measurement will be used in the analysis (as done currently). 

This method is based on the Timepix3 calibration, and many parts of the code are directly copied from here: https://github.com/eudaq/eudaq/blob/master/user/timepix3/module/src/Timepix3Event2StdEventConverter.cc

cc: @ffeindt

### DIFF
--- a/user/caribou/module/src/H2MEventConverter.cc
+++ b/user/caribou/module/src/H2MEventConverter.cc
@@ -7,6 +7,7 @@
 
 #include <string>
 #include <algorithm>
+#include <cmath>
 
 using namespace eudaq;
 
@@ -15,6 +16,8 @@ namespace {
       H2MEvent2StdEventConverter>(H2MEvent2StdEventConverter::m_id_factory);
 }
 
+static bool m_first_time = true;
+static std::vector<std::vector<float>> vtot; // for ToT calibration
 
 bool H2MEvent2StdEventConverter::Converting(
     eudaq::EventSPC d1, eudaq::StandardEventSP d2,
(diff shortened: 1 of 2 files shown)


user/caribou/module/src/H2MEventConverter.cc
```diff
@@ -143,3 +185,42 @@ bool H2MEvent2StdEventConverter::Converting(
...
+        if(line.size() > 0 && isdigit(line.at(0))) {
+            std::stringstream ss(line);
+            std::string word;
+            std::vector<float> row;
+            while(std::getline(ss, word, delim)) {
+                i += 1;
+                row.push_back(stof(word));
+            }
+            dat.push_back(row);
+        }
```
simonspa, line 217: I would also check the inner vectors to have the correct size

user/caribou/module/src/H2MEventConverter.cc
```diff
@@ -119,8 +135,34 @@ bool H2MEvent2StdEventConverter::Converting(
...
 
+    // best guess for charge is ToT if no calibration is available
+    EUDAQ_DEBUG("tot is: " + to_string(tot));
+
+    double charge = static_cast<float>(tot);
+    if(!vtot.empty()) {
+        EUDAQ_DEBUG("Applying calibration to DUT");
+        size_t scol = static_cast<size_t>(col);
+        size_t srow = static_cast<size_t>(row);
+        float a = vtot.at(64 * srow + scol).at(2);
```
simonspa, line 146: what's stored in `vtot.at(X).at(0)` and `vtot.at(X).at(1)`? If it's just the coordinates, I would drop them when reading the file since you don't need them

simonspa (commented):
Looking good! Two small comments

ffeindt:
Not much to add, but in order to prevent a mess, maybe we should merge https://github.com/eudaq/eudaq/pull/750 first and rebase?

user/caribou/module/src/H2MEventConverter.cc
```diff
@@ -143,3 +185,49 @@ bool H2MEvent2StdEventConverter::Converting(
...
+        // if not write to output vector
+
+        if(line.size() > 0 && isdigit(line.at(0))) {
+            std::stringstream ss(line);
+            std::string word;
+            std::vector<float> row;
+            for (size_t i = 0; i < 6; ++i) {
+                std::getline(ss, word, delim);
+                if (i >= 2) { // Keep only calibration data
+                    row.push_back(std::stof(word));
```
simonspa, line 216: Tiny thing, and just because I wanted to mention this - now you discard the addresses but never check them. Imagine (for some arcane reason):

```
0 1 9 9 9 9
0 3 9 9 9 8
0 2 9 9 9 7
```

would mean that you accidentally swap the values of pixel (0, 2) and (0, 3) because you assume they are in the file in the correct order.

...but let's not get overboard with nitpicky here. :wink:

simonspa (approved):

PR #767, opened 2024-12-10
Fill euCliReader stat flag with a little bit of life

annkvth:
euCliReader has an option flag `"s", "statistics"`  that promises to `"enable print of statistics"`  but currently does nothing (it's not even checked). 

This commit makes it output at least the trigger number range and a list of devices found in the data.

Example output:
example-output


arummler (approved):
Looks okay and scope is limited. CI failure due to unrelated issue (adenium, MR open).

PR #772, opened 2025-02-09
Fix NOMINMAX for windows

arummler:
euCliReader did not compile anymore on Windows. Not clear where windows.h gets pulled in now? Could be the boost from CMS? But should be isolated. This really should be cleaned up but it is a fix for the moment. Took the opportunity to clean up also some outdated root stuff.


lhuth (approved):
lg2m

PR #777, opened 2025-03-14
naming the folder after the run number

lhuth:
@kraemeru 
Set the run number as folder name - we can extend this to whatever we need if we want


simonspa (approved):
Fodlers should *always* be named after the run number. :wink:

PR #780, opened 2025-03-22
Modernize StdEvtMonitor CMake

arummler:
- target based cmake handling of ROOT linking
- added template specialization of from_string<bool> utility function which made it fail on Windows
- remove ROOT5 compatibility

### DIFF
--- a/main/lib/core/include/eudaq/Utils.hh
+++ b/main/lib/core/include/eudaq/Utils.hh
@@ -165,6 +165,9 @@ namespace eudaq {
   from_string(const std::string &x, const uint16_t &def) {
     return static_cast<uint16_t>(from_string(x, (uint64_t)def));
   }
+  
+  template <>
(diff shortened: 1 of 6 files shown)


arummler:
@lhuth That's the next fix/upgrade I would like to merge.

arummler:
@lhuth @herkerta Can we proceed with this one? If you think that ROOT5 still needs to be supported I can take it out but the option bug fix is important.

lhuth:
I think root5 can be considered deprecated and we do. It need to further support it. Will look at the rest next week. Am 26.04.2025 um 14:13 schrieb Andre Rummler ***@***.***>:﻿
arummler left a comment (eudaq/eudaq#780)
@lhuth @herkerta Can we proceed with this one? If you think that ROOT5 still needs to be supported I can take it out but the option bug fix is important.

lhuth (approved):

lhuth:
tested by @herkerta and no issues seen

PR #783, opened 2025-03-23
Add CI integration for new Corryvreckan monitor

arummler:
Enabling CI:
- XROOTD: linux gcc, linux clang, mac
- CORRYMON: linux gcc, linux clang


arummler:
I will merge it as it concerns only the CI.

PR #787, opened 2025-04-10
TTree CMAKE update / enable on Windows

arummler:


arummler:
Based on #780 

@lhuth Second MR to be considered after #780

PR #788, opened 2025-04-11
ROOT monitor cmake overhaul and activation on Windows

arummler:


herkerta (approved):
Looks fine to me. But I have to say I never use the ROOT monitor and I don't think anyone at DESY ever does. So, I didn't run it as a check.

PR #789, opened 2025-05-13
Add producer for Keithley sourcemeter (2450 / 2470)

annkvth:
For our last testbeam, I wrote a Producer to steer our power supplies, so that we could e.g. do bias scans via the EUDAQ2 scan feature. 
This might be useful again in the future and/or for others, hence this pull request.

P.S. I'm happy to change the subfolder name if someone has a nicer suggestion.


lhuth (approved):
User code - usage without warranty given

PR #795, opened 2025-07-05
Fix core build target and depending header paths

arummler:
In the build path of the eudaq_core target the /include is missing which makes inclusion in down stream projects impossible without workarounds. Fixing also the paths after removing the non-target include. (There are still more cmake style violations to be fixed but trying to keep in concise.)


arummler:
Small fix of cmake (needs a bunch of others in other places). Ran into it when removing build workarounds In YarrEudaqProducer which consumes eudaq directly. It should not have any side effects for anybody who is not doing target_link_libraries to eudaq_core. And it is anyhow broken.

herkerta (approved):

PR #808, opened 2026-04-24
fix build of manual (latex)

arummler:
There were issues with the encoding as well as the listing package. Also a bit of modernization (package wise).


herkerta (approved):
I didn't see the breaking error but the changes look ok anyway.

PR #823, opened 2026-08-20
Eudet  fix spurious rollovers

ffeindt:
Fixing spurious trigger counter rollovers caused by faulty trigger IDs in EUDED data.
I did not find a nice way to handle this separately for all planes. Better Ideas?
Tested on data from the ongoing SPARC test beam.

### DIFF
--- a/user/eudet/module/src/NiRawEvent2StdEventConverter.cc
+++ b/user/eudet/module/src/NiRawEvent2StdEventConverter.cc
@@ -127,17 +127,19 @@ bool NiRawEvent2StdEventConverter::Converting(eudaq::EventSPC d1, eudaq::Standar
 
       // if we see a jump and the two conditions are met, correction might have been spurious
       if(abs_difference > 1 && condition_this && condition_previous) {
-        EUDAQ_WARN("            Detected spurious rollover in event: " + std::to_string(d1->GetTriggerN()));
-        EUDAQ_WARN("                                 previous event: " + std::to_string(m_previous_trigger_id));
+        EUDAQ_WARN("Detected spurious rollover in event: " + std::to_string(d1->GetTriggerN()));
+        EUDAQ_WARN("                     previous event: " + std::to_string(m_previous_trigger_id));
         if(m_last_corrected_ID != d1->GetTriggerN()){
           m_spurious_rollovers++;
           // can not keep track of plane ID 
           // need to check if this was already corrected
           m_last_corrected_ID = d1->GetTriggerN();
-          EUDAQ_WARN("                rollover counter incremented to: " + std::to_string(m_spurious_rollovers));
+          EUDAQ_WARN("    rollover counter incremented to: " + std::to_string(m_spurious_rollovers));
+          EUDAQ_WARN("Be weary and check synchonization!");
+
         }
         else{
-          EUDAQ_WARN(" rollover counter already incremented for event: " + std::to_string(m_last_corrected_ID));
+          EUDAQ_WARN("Event already corrected.");
         }
         
         return false;
--- a/user/eudet/module/src/NiRawEvent2StdEventConverter.cc
+++ b/user/eudet/module/src/NiRawEvent2StdEventConverter.cc
@@ -26,7 +26,7 @@ namespace{
 }
 
 bool NiRawEvent2StdEventConverter::m_configured = false;
-bool NiRawEvent2StdEventConverter::m_correct_spurious_rollovers;
+bool NiRawEvent2StdEventConverter::m_correct_spurious_rollovers=false;
 uint64_t NiRawEvent2StdEventConverter::m_previous_trigger_id=0;
 uint64_t NiRawEvent2StdEventConverter::m_spurious_rollovers=0;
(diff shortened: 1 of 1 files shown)


simonspa:
I think you are handling it for all planes, since they are all within the same event in different data blocks.
Our current decoder is inefficient as in from the Corry side, the same event will be decoded 6 times, once for each plane. So we do the correction 6 times independently.

All fine from my view!

user/eudet/module/src/NiRawEvent2StdEventConverter.cc
```diff
@@ -99,6 +116,46 @@ bool NiRawEvent2StdEventConverter::Converting(eudaq::EventSPC d1, eudaq::Standar
...
+      bool condition_this = (0x7fff & d1->GetTriggerN()) < 0x2000;
+      bool condition_previous = (0x7fff & m_previous_trigger_id) > 0x6000;
+
+      // if we see a jump and the two conditions are met, correction might have been spurious
+      if(abs_difference > 1 && condition_this && condition_previous) {
+        EUDAQ_WARN("Detected spurious rollover in event: " + std::to_string(d1->GetTriggerN()));
+        EUDAQ_WARN("                     previous event: " + std::to_string(m_previous_trigger_id));
+        if(m_last_corrected_ID != d1->GetTriggerN()){
+          m_spurious_rollovers++;
+          // can not keep track of plane ID
```
simonspa: Drop this, other planes are decoded separately, or am I mistaken? Do these calls (from other module instances) land in the same static decoder instance?

(arrrrrrrg!)

user/eudet/module/src/NiRawEvent2StdEventConverter.cc
```diff
@@ -12,16 +12,34 @@ class NiRawEvent2StdEventConverter: public eudaq::StdEventConverter{
...
+  static uint64_t m_last_corrected_ID;
 };
 
 namespace{
   auto dummy0 = eudaq::Factory<eudaq::StdEventConverter>::
     Register<NiRawEvent2StdEventConverter>(NiRawEvent2StdEventConverter::m_id_factory);
 }
 
+bool NiRawEvent2StdEventConverter::m_configured = false;
+bool NiRawEvent2StdEventConverter::m_correct_spurious_rollovers;
```
simonspa:
```suggestion
bool NiRawEvent2StdEventConverter::m_correct_spurious_rollovers=false;
```

simonspa:
I see now. Probably the last-id check is an adequate measure.

simonspa (dismissed):

user/eudet/module/src/NiRawEvent2StdEventConverter.cc
ffeindt: I tested a different version, without the last ID check, and it ended up incrementing 6 times (once for each plane) per rollover. I meant to catch this using an array with plane ID as index, but the `id` variable in the code is `0` for all planes. I did not find anything else I can use to separate the planes.

simonspa: Now this only works because all 6 planes progress at the same time to the next trigger, otherwise also previous ones would be incremented as soon as the first plane ran into the issue.

In principle the IDs should count up, looking at the code, since `board` is incremented in the loop and `m_ids` will then return 0->5. *but* for each call of the decoder, *all* planes are decoded, no matter which one you are using in the end and I think this is where the issue comes from.

herkerta: I think initializing m_last_corrected_ID=0 leads to an issue when the (first) spurious trigger ID is 0 and condition_previous==true. (The general case of spurious trigger ID == m_last_corrected_ID should be unlikely enough.)

ffeindt: So you say the decoder is called 6 times, and each time we decode and return all planes and select the corresponding plane in the EvenLoader? But how comes that the warnings are printed only 6 times then?

Addition: ahh, because abs_difference is 0 then.

simonspa: Have to look again how we cache decoded events in the loader :thinking:

ffeindt: and for Adrians comment: I see... I thought this works, because we have the case in the data..., but we might have gotten lucky with the condition on the previous trigger ID. It should work if I initialize it to `std::numeric_limits<uint64_t>::max()`, right?

simonspa: make it signed and initialize to `-1`

herkerta: I vote for std::numeric_limits<uint64_t>::max().

ffeindt: Simon and I discussed a little more, and since the plane index is not used, the thing can be pulled out of the while loop. But thats something for next week ;)

herkerta (approved):

